PR #25, opened 2019-03-04
yearmonthdayhour function

Balinus:
This adds the `yearmonthdayhour` function.

Not sure where to put tests though. I'll update the PR if you can point it out or if I find it.

edit - Also! There is a small typo in the title of the package "[...] NetCDF files in **JuIlia**" -> "[...] NetCDF files in **JuIia**"

### DIFF
--- a/src/time.jl
+++ b/src/time.jl
@@ -4,12 +4,12 @@ if VERSION >= v"0.7.0-beta.0"
     using Dates
     import Dates: UTInstant, Millisecond
     import Dates: year,  month,  day, hour, minute, second, millisecond
-    import Dates: daysinmonth, daysinyear, yearmonthday, yearmonth, yearmonthdayhour
+    import Dates: daysinmonth, daysinyear, yearmonthday, yearmonth
     import Dates: monthday, len
 else
     import Base.Dates: UTInstant, Millisecond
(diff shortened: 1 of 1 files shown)


Balinus:
Just understood that `yearmonthdayhour` is not a `Dates` function that we are extending. So perhaps the place of such function is more relevant directly in `ClimateTools`. It's up to you!

Alexander-Barth:
I indeed try to limit the scope of the time function to what is already in julia Base.
In my opinion it would be more appropriate to have these functions in your ClimateTool package (as you need to define them for the regular `DateTime` objects too).

I think that you can implement this function as following:

```
julia> yearmonthdayhour(dt::DT) where DT <: Dates.TimeType = (Dates.year(dt),Dates.month(dt), Dates.day(dt), Dates.hour(dt))
yearmonthdayhour (generic function with 1 method)

julia> yearmonthdayhour(DateTime(2000))
(2000, 1, 1, 0)

julia> yearmonthdayhour(DateTimeNoLeap(2000))
(2000, 1, 1, 0)
```
This will work for `DateTime` and the additional time types in NCDatasets. 

Thanks a lot for spotting the typo in the description!!

Balinus:
A pleasure!

Ok, thanks for the generic implementation! I'll put it directly in ClimateTools.

PR #35, opened 2019-08-07
dimensions

rafaqz:
This is a proof of concept than a serious pull request at this stage, there will be bugs to iron out first and some tests to write.

Its only a few lines of code but the (small) [GeoData.jl](https://github.com/rafaqz/GeoData.jl/) and its [DimensionData.jl)](https://github.com/rafaqz/DimensionalData.jl/ dependencies add recipes for labelled plotting, indexing and selection by dimensions, reduction over dimension with mean, reduce etc, and tracking of metadata through all these transformations.
```julia
# first time install
# ] dev http://github.com/rafaqz/CoordinateReferenceSystemsBase.jl
# ] dev http://github.com/rafaqz/DimensionalData.jl
# ] dev http://github.com/rafaqz/GeoData.jl

using NCDatasets, GeoData, Statistics, Plots
# download a test file
filename = "tos_O1_2001-2002.nc"
download("https://www.unidata.ucar.edu/software/netcdf/examples/$filename", filename)
ds = Dataset(filename)
v = ds["tos"];
# Plot the third month
v[Time(3)] |> plot
# Move data to a memory backed GeoArray
g = v[:,:,:]
# Plot the mean sea surface temperature # around Australia from Aug to Dec 2002:
dimranges = Time((DateTime360Day(2002, 08, 1), DateTime360Day(2002, 012, 30))), 
            Lat(-45:0.5), 
            Lon(110:160)
select(g, dimranges) |> x->mean(x; dims=Time) |> plot
close(ds)
```
It would be good to be able to save these back to NetCDF automatically, which will be fairly straightforward as the arrays keep all the metadata and update the dims as you change things. Let me know what you think.


Alexander-Barth:
This is indeed interesting work, but can this be implemented in a separate package on top of NCDatasets? Also it would make it more simpler to support other raster formats (like [Zarr](https://zarr.readthedocs.io/en/stable/spec/v2.html)) in the future.

```julia
# Are these dimension names actually canonical?
const dimmap = Dict("lat" => GeoData.Lat, "lon" => GeoData.Lon, "time" => GeoData.Time) 
```
No unfortunately not. Some grid are also rotated (e.g. https://www.myroms.org/forum/viewtopic.php?f=1&t=850  dimensions do not align with longitude and latitude in this case)  or completely unstructured. Only the attribute standard_name of the variable is "standard".

rafaqz:
I've also come around to thinking it would be better as a series of thin wrapper packages between various data formats and GeoData.jl (ie including one for Zarr etc). It would also mean I can just release them incomplete (95% of use cases) where that would be bad here. GeoData.jl will eventually support rotation and two-dimension lookups using wrapper types and affine maps or matrices inside the dims, but maybe not for a while as I don't need it very often.

And yes it's unfortunate CF didn't enforce a naming standard, but those few common names work with most of their test data. The fallback is Dim{:somedim}(x) if it can't find a common one, its just an uglier syntax to index with than `Lat(x), Lon(y)`.

PR #44, opened 2019-12-20
Improve README, add TOC, add alias NCDataset

Datseris:
With this PR I am hoping to improve the README a bit by adding a table of contents, re-arranging things for the docs to be a bit more intuitive and adding slightly more info here and there.

I am also adding the `NCDataset` alias (if you approve) for two reasons:
1. In DynamicalSystems.jl we also use the same name, `Dataset` for a central structure.
2. Because it is typical that a julia package "ThisThings.jl" would define a central structure `ThisThing` (without final s).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
 keywords = ["netcdf", "climate and forecast conventions", "oceanography", "meteorology", "climatology", "opendap"]
 license = "MIT"
 desc = "Load and create NetCDF files in Julia"
-version = "0.9.5"
+version = "0.9.6"
 
 [deps]
 BinDeps = "9e28174c-4ba2-5203-b857-d8d62c4213ee"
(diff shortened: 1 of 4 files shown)


Datseris:
See https://julialang.github.io/Pkg.jl/v1/creating-packages/index.html , regarding pluralized names.

Datseris:
(actually according to this style guideline there should only be `NCDataset` , but of course you may not want to go forward with such a hugely breaking change)

Datseris:
Well, coveralls will complain no matter what I'll do, so I'll leave it as is.

Alexander-Barth:
Don't worry about coverage, it's just 0.1 %
I am currently away from my laptop. What do you think, how likely is PR to break user code? 
Thanks!

Datseris:
> What do you think, how likely is PR to break user code? 

Exactly zero! :D  There is no change in code, I've only added the alias `NCDataset` (which means that besides `Dataset` you can use `NCDataset`). Everything else is documentation improvements.

Alexander-Barth:
Ok, thank you for your contribution!

Datseris:
> Ok, thank you for your contribution!

Oh you are welcome, and there are many more to come (see here https://discourse.julialang.org/t/newcomer-contributor-in-juliageo-and-co-help-me-get-started/32480/7 )

PR #53, opened 2020-01-07
nsize function for CFVariable

### LINKED ISSUE #52: Make `size` of a NCDataset/CFVariable return a NamedTuple instead
I think it would be super-useful to have `size` of some `CFVariable` return a named tuple instead of just a tuple, like so:

Current
```
julia> fld
  Datatype:    Float32
  Dimensions:  lon × lat × time
  Attributes:
   comment              = none
   units                = W m-2
   _FillValue           = -999.0

julia> size(fld)
(360, 180, 228)
```

proposed:
```
julia> size(fld)
(lon = 360, lat = 180, time = 228)
```
this is non-breaking, as the result can still be accessed with integer indices. But it can be also accessed with name, which is super intuitive and gets rid of the problems "which dimension is which".

Datseris:
Closes #52

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -157,6 +157,7 @@ path
 defVar
 dimnames
 name
+nsize
 chunking
 deflate
 checksum
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -1328,6 +1328,18 @@ Dataset(var::CFVariable) = Dataset(var.var)
 
 Base.size(v::CFVariable) = size(v.var)
 dimnames(v::CFVariable)  = dimnames(v.var)
+
+"""
+    nsize(v::CFVariable)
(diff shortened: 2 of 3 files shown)


Alexander-Barth:
Thanks for your PR. Do you know of any other package/array type implementing such a function? We might consider to use the same name for the function. In any case, I actually do prefer to not reuse `size`.

Datseris:
I don't know of any, but naming is very easy to change and deprecate in favor of a better name. If you have a better suggestion go ahead (I actually like `nsize`).

Datseris:
If you don't like `nsize` we could do `namedsize` or `dimensionsize` or `dimensions` (which would also be defined for an `NCDataset` directly).

Balinus:
How about `dimsize()`?

Alexander-Barth:
I like `dimsize` too! Thanks to both for you!

PR #55, opened 2020-01-07
Big documentation re-structure and improvement!!!

Datseris:
Quick comment: I think we should rethink the syntax `v[:]`, or hide it from the users... It makes little sense that `v[:]` doesn't return `Vector`, when literally every other operation on `v` behaves as if `v` was an `Array`.

So, with this PR I think I am improving the documentation considerably by adding _a lot_ more text for newcomers and separating into pages!


Datseris:
As the git diff is quite massive, the best way to review this is probably to build the docs locally...

Alexander-Barth:
This is great. Thanks!

PR #58, opened 2020-01-09
WIP: SubArray/view/select

### LINKED ISSUE #47: Convenience selection function for sub-parts of NCDataset
Hi there, I have a feature request which could be helpful, albeit it is a small convenience syntax.

Let's say I have loaded an NCDataset. In Python this is done with `xarray`, but here we have something like
```
ebaf41_toa = NCDataset(datadir("CERES", "CERES_EBAF-TOA_Ed4.1_Subset_200003-201902.nc"))
```

Now, let's say we have a field of this dataset, `fld = ebaf41_toa["toa_sw_all_mon"]`:
```
  Datatype:    Float32
  Dimensions:  lon × lat × time
  Attributes:
   comment              = none
   units                = W m-2
   _FillValue           = -999.0
```

In Python, you would do something like
```
ebaf41_toa = xr.open_dataset(path)
fld = ebaf41_toa.toa_sw_all_mon
```

This `xarray` offers a convenience syntax like 
```
fld.sel(time=slice('2000-03-01','2019-03-01'))
```
which means that you can select sub-parts of the field by specifying which ranges of the dependent variables to keep. 

This could be implemented here as well, but one has to somehow map the given keywords to "which" dependent variables they represent. I am happy to make this contribution if you are willing to guide me.

Datseris:
Closes #47

Todo:

- [ ] Group variable and co under common abstract type
- [ ] Create a `SubArray` type for Variable and co. that stores a limit / indexing array for each of the `dimension`s of the Variable. This will be the guiding principle for this type: https://docs.julialang.org/en/v1/devdocs/subarrays/#SubArrays-1 
- [ ] Implement `view` that returns this subarray
- [ ] Implement `select` that uses the names to make a `view`
- [ ] Tests
- [ ] Docs

### DIFF
--- a/src/select.jl
+++ b/src/select.jl
@@ -1,0 +1,28 @@
+export select
+
+
+"""
+    select(v::Variable; dimensions_to_limits...)
+Select the subset of the variable `v` by keeping the values of
+each of the given dimension that are between the corresponding limits.
+Return a _view_ into `v`, which retains its type (i.e. again a `Variable`,
+not an `Array` like `getindex`.
+
+This function is called like
+```julia
+select(v; lat = (-90, 0), lon = (0, 180))
+```
+each keyword denotes a dimension to index into, and select values of `v` along
+that dimension that satisfy `lower ≤ value ≤ upper`. Notice that for
+`select` the values given to each keyword represent true numeric values of
+the dimensions and not integers (see [`iselect`](@ref) for that).
+There for you could also use syntax like `time = (Date("2015"), Date("2016"))`
+if one of your dimensions has a value that is `Date`.
+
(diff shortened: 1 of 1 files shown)


Datseris:
@Alexander-Barth I must admit, the way the source code is structured is a bit unwieldy: the main file `NCDatasets.jl` is several thousand lines long, which does not make it immediatelly clear where in the source code is each thing implemented. Is it okay if I try to organize the source a bit in this PR? Or should I open a new PR before I start working on this to try to organize?

Datseris:
Closing this in favor of adding transformation from `CFVariable` to `AxisArray`, which would do everything for us for free.

PR #59, opened 2020-01-09
Seperate source code into files with sections

Datseris:
Ongoing work to allow more people to more easily read, understand, and contribute back to NCDatasets.jl


Datseris:
I am a bit confused about the parts related to the NetCDF library: I thought it is a C library, yet for some reason NCDatasets.jl uses Conda to install it. Why?

Datseris:
Another point is removing or not the commented lines in the `netcdf_c.jl`: do they serve any purpose so that they stay here?

Balinus:
> I am a bit confused about the parts related to the NetCDF library: I thought it is a C library, yet for some reason NCDatasets.jl uses Conda to install it. Why?

I think it's simply a matter of having an easy mean to install `libnetcdf`: it is available through CondaBinDeps. The build process check if there is an adequate system version before using conda.

Datseris:
But `Conda` requires a full python installation via miniconda, right? Shouldn't we try to leverage the numerous advances of `Pkg` in installing binary dependencies to make this process more Julian?

Balinus:
> But `Conda` requires a full python installation via miniconda, right? Shouldn't we try to leverage the numerous advances of `Pkg` in installing binary dependencies to make this process more Julian?

There's BinaryProvider.jl that would be adequate in that respect. 

With that being said, I find that those kind of modifications are quick to do, but long before it works flawlessly. Hence, I'd go with the popular adage: "if it's not broken, don't fix it!".

Datseris:
@Alexander-Barth I checked the code again, and I am not sure how to organize things further. In any case, I am requesting a review here. If you think something can be made even more clear let me know and I'll do it. I'd like this to be merged first before working on #58 .

Alexander-Barth:
Thanks for your work. There are two binary files
`test/temp1.nc` and `test/test_cont_ragged_array.nc`. Is this an error or was this intentional?

Alexander-Barth:
> I am a bit confused about the parts related to the NetCDF library: I thought it is a C library, yet for some reason NCDatasets.jl uses Conda to install it. Why?

It would be really great if we could drop the conda dependencies, but I did not succeed to do this. The biggest problem is cross-compilation of HDF5 (and then NetCDF).

https://github.com/JuliaPackaging/Yggdrasil/issues/329

Datseris:
> Thanks for your work. There are two binary files
test/temp1.nc and test/test_cont_ragged_array.nc. Is this an error or was this intentional?

Thanks for pointing this out. This is not intentional; I guess it happened because I've run the tests. But it made us identify that some tests do not "clean up" after themselves and leave created files. I'll try to identify them and add `rm` in the tests.

Datseris:
@Alexander-Barth I've fixed the test issues, you can merge if you see it fit.

PR #62, opened 2020-01-10
Add comment in Manual pointing explicitly to the docs

Datseris:
I realized that having such a well-written README can be a double-edged knife, as some users might immediately start reading the manual and forego the actual docs (I actually did this in my first read through NCDatasets.jl).

This removes this danger.


Alexander-Barth:
Thanks, I agree.

PR #64, opened 2020-01-20
Improve docs: mention the `mean` problem

### LINKED ISSUE #48: Something is wrong with `mean`: takes too much time
**Describe the bug**

Calling `mean` on a NCDataset seems to take too much time. Much more than when applied to the direct array representation.

**To Reproduce**
Let's say I have:
```
julia> fld
  Datatype:    Float32
  Dimensions:  lon × lat × time
  Attributes:
   comment              = none
   units                = W m-2
   _FillValue           = -999.0

julia> fldarray = Array(fld);
```
Then I do:
```
julia> mean(fldarray)
102.24223f0
```
which is instant. However, 
```
julia> mean(fld)
```
takes way too long (I always have to terminate Julia process). The same story goes for using `dim(fld; dims = 3)`...

**Expected behavior**

`mean` should take almost as much time as in the Array version.

Datseris:
Closes #48

I am still in the process of understanding GeoData.jl. DimensionalData.jl I have understood and started contributing already, but GeoData.jl is much more complicated so it will take some time until I can mention it here in a "user-friendly way".

Until then though, I think it is important to have this mention in the docs, as it will prevent problems like #48.

I am also thinking of the following, and please Alexander give me your feedback: we should _remove_ the documentation from the README, as it has two dangers:

1. Makes it harder to maintain, as a change in the software must be translated into **two** changes in the documentation.
2. It gives users the possibility of never reading the actual docs, which have many useful info such as this one.

What do you think?


Alexander-Barth:
The mean problem is already in the docs under performance tips:

* Most julia functions (like `mean`, `sum`,... from the module Statistics) access an array element-wise. It is generally much faster to load the data in memory (if possible) to make the computation.

```julia
using NCDatasets, BenchmarkTools, Statistics
ds = NCDataset("file.nc","c")
data = randn(100,100);
defVar(ds,"myvar",data,("lon","lat"))
close(ds)

ds = NCDataset("file.nc")
@btime mean(ds["myvar"]) # takes 107.357 ms
@btime mean(ds["myvar"][:,:]) # takes 106.873 μs, 1000 times faster
close(ds)
```

Actually, I like when projects include some kind of Quick Guide in the README.md which in case of NCDatasets.jl is indeed ~80% of what you need to know (IMHO).

Datseris:
Oh damn, how could I forget this, _I_ wrote it :P 

Okay for the readme. Should we still though keep this advice in the "Variables" page as well? Makes it more easy to spot and it isn't so much extra text.

Alexander-Barth:
OK, I shorten it a bit and referrer to Performance tips for me information.

PR #66, opened 2020-01-21
Improve `nomissing` functionality

Datseris:
1. Improves performance of searching for missings in `da`.
2. Allows the user to pass in reshaped arrays, which are e.g. the result of `dropdims` function, after one has e.g. taken a mean across a dimension.


Alexander-Barth:
Thanks, George!

PR #67, opened 2020-01-22
bump CondaBinDeps dependency, bump patch version

dkarrasch:
I wonder if your package really depends on `Printf` and `Random`, or whether these are only used in the tests. For test dependencies, you have the `[extras]` section, anyway, so maybe you could reduce the amount of declared package dependencies.


Alexander-Barth:
@dkarrasch Thank for having an eye on the version of the dependency! Actually, I enabled CompatHelp, and I got already this PR https://github.com/Alexander-Barth/NCDatasets.jl/pull/68 which is already merged. Thank you also for noticing the potential issue about Printf and Random. I will do a test to see if I can remove them.

dkarrasch:
No problem. So I was faster than the bot! 😄  🎉

PR #81, opened 2020-04-13
anti-alias doc

visr:


Alexander-Barth:
:-)

(I guess I replaced all Dataset by NCDataset too quickly)
Thanks!

PR #93, opened 2020-08-04
do not throw when closing closed NCDataset

visr:
This copies the behavior of `Base.close` on file handles:

```julia
julia> io = open("asd","w")
IOStream(<file asd>)

julia> close(io)

julia> close(io)

julia>
```

And is more convenient for interactive use if you can't quite remember if you already closed a `NCDataset`.

The first commit fixes an issue in the usage of `rethrow`, though I guess this was rarely triggered.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -80,6 +80,6 @@ function Base.show(io::IO, a::BaseAttributes; indent = "  ")
                 return
             end
         end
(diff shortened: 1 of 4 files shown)


Alexander-Barth:
Thanks a lot, @visr ! I will make a new release soon with this.

PR #111, opened 2020-11-14
Update performance.md

aramirezreyes:
Added a note saying that load! is not exported


Alexander-Barth:
Thanks @aramirezreyes !

PR #113, opened 2020-11-22
use Ref instead of size 1 Vector

visr:
Ran into this at #88 and update 1 in e02929697b71dc9add90bf57673932829f987504. This updates all the others. Interestingly it looks like this change actually exposed a few existing issues.

### DIFF
--- a/src/netcdf_c.jl
+++ b/src/netcdf_c.jl
@@ -464,8 +464,8 @@ function nc_inq_vlen(ncid::Integer,xtype::Integer)
     return unsafe_string(pointer(name)),datum_sizep[],base_nc_typep[]
 end
 
-function nc_free_vlen(vl)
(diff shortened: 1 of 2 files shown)


Alexander-Barth:
Thanks for this PR. If I see correctly, your  PR works on Julia 1.5 and nightly but fails on 1.0. However, using NetCDF_jll will require julia 1.3 anyway.

visr:
Yes indeed, I just did a quick rebase on top of the new master, expecting tests to pass now.

visr:
Do you want to retain Appveyor and Travis build besides the Github Actions by the way? I notice they are much slower.

Alexander-Barth:
Indeed, I will remove Travis. But I plan to keep Appveyor as I don't know who will be the next who will charge for CI :-)
Thanks for the re-base!

PR #119, opened 2021-01-15
Change attribute color from blue to cyan

briochemc:
This is just a suggestion, but I think cyan works better with dark backgrounds (blue is almost unreadable for me sometimes), while it still works with lighter/white backgrounds. 

Here are some screenshots of different backgrounds (I mostly use the last one, where the blue output is really hard to read for me):

Screen Shot 2021-01-15 at 11 51 52 am
Screen Shot 2021-01-15 at 11 52 27 am
Screen Shot 2021-01-15 at 11 53 11 am

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -70,7 +70,7 @@ function Base.show(io::IO, a::BaseAttributes; indent = "  ")
         # use the same order of attributes than in the NetCDF file
         for (attname,attval) in a
             print(io,indent,@sprintf("%-20s = ",attname))
-            printstyled(io, @sprintf("%s",attval),color=:blue)
+            printstyled(io, @sprintf("%s",attval),color=:cyan)
             print(io,"\n")
         end
     catch err


Alexander-Barth:
I agree, thanks @briochemc, cyan is a better default. For the record, here is how julia handles color for e.g. @info:

https://github.com/JuliaLang/julia/blob/985dfa54c333e58d40c1b28de7692dc2cce4c965/base/client.jl#L24

It is also cyan but you can override it with an environment variable. Would that be useful also here?

briochemc:
Ahah! Good find! I was looking for that part but did not know where to look so I eventually gave up and went with `:cyan`... 

Just to be sure we are aligned, you are saying we should use
```julia
printstyled(io, @sprintf("%s",attval),color=Base.info_color())
```
instead? And apply the same reasoning to the other calls to `printstyled`?

My hope is that this way anyone overriding the REPL colorscheme (e.g., OhMyREPL.jl) would see things in his preferred colors when using NCDatasets as well? (I'm actually not sure how this all works, even looking into OhMyREPL for example.)

Alexander-Barth:
We can keep it simple for now and hard-code cyan. But if somebody comes along and prefers blue, I think it would be better to have it configurable using e.g. a environement variable `NCDATASET_ATTRIB_COLOR` as `info_color` is not exported from Base.

briochemc:
So having
```julia
const default_color_attribute = :cyan
attribute_color() = Base.repl_color("NCDATASET_ATTRIB_COLOR" , default_color_attribute)
```
and then replace the only edited line of this PR into 
```julia
printstyled(io, @sprintf("%s",attval),color=attribute_color())
```
?

Alexander-Barth:
OK, thanks a lot! I will merge it and update it in the master branch.

briochemc:
Oh wait, I thought you wanted the `attribute_color()` function instead? Let me know and I can make another PR with those changes if you want :)

Alexander-Barth:
No problem, I copied your code here:

https://github.com/Alexander-Barth/NCDatasets.jl/commit/ccdfeca1a7f2dbe6152e07bbb624b1079a1b542c

We already now have a attribute_color() function :-) Other colors are configurable as well.

Issue #120, opened 2021-01-26
Missing method for NCDatasets.load!(ds["var"],buf,:,1)


aramirezreyes:
**Describe the bug**

Trying to load in-place data using a mixture of ranges and integers fails.

**To Reproduce**

```
NCDataset("testfile.nc","c") do ds
      data = Int32[i+3*j for i = 1:5, j = 1:6]
      defVar(ds,"temp",data,("lon","lat"), attrib = [
          "units" => "degree_Celsius",
          "long_name" => "Temperature"
      ])
  
  data2 = zeros(eltype(data),5,1)
      NCDatasets.load!(ds["temp"].var,data2,:,1)
      @show data2 == data[:,1]

rm("testfile.nc")

end
```

**Expected behavior**

Data should be loaded 

**Environment**

 - macOS Big Sur
 - Julia 1.5.1
 -NCDatasets v0.11.3
 
**Full output**

```
 NCDataset("testfile.nc","c") do ds
             data = Int32[i+3*j for i = 1:5, j = 1:6]
             defVar(ds,"temp",data,("lon","lat"), attrib = [
                 "units" => "degree_Celsius",
                 "long_name" => "Temperature"
             ])

         data2 = zeros(eltype(data),5,1)
             NCDatasets.load!(ds["temp"].var,data2,:,1)
             @show data2 == data[:,1]

       rm("testfile.nc")

       end
```

```
ERROR: MethodError: no method matching ncsub(::Tuple{Base.Slice{Base.OneTo{Int64}},Int64})
Closest candidates are:
  ncsub(::Tuple{Vararg{T,N}}) where {T, N} at /Users/arreyes/.julia/dev/NCDatasets/src/variable.jl:372
Stacktrace:
 [1] load!(::NCDatasets.Variable{Int32,2,NCDataset}, ::Array{Int32,2}, ::Colon, ::Int64) at /Users/arreyes/.julia/dev/NCDatasets/src/variable.jl:121
 [2] (::var"#9#10")(::NCDataset{Nothing}) at ./REPL[18]:9
 [3] NCDataset(::var"#9#10", ::String, ::Vararg{String,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/arreyes/.julia/dev/NCDatasets/src/dataset.jl:286
[...]

PR #121, opened 2021-01-26
Adding method to load in-place mixing ranges and ints as indices.

aramirezreyes:
This is work in progress regarding #120, currently, this commit creates a test that reproduces the error


aramirezreyes:
This fix makes my test work and it also doesn't break any other tests (apparently). Would appreciate review.

Alexander-Barth:
This is great! Fix and test code, every thing is there. Thank you so much!

PR #123, opened 2021-02-16
verify1.6-nightly tests

visr:
On NetCDF.jl Windows CI is failing with an unclear error: 
https://github.com/JuliaGeo/NetCDF.jl/pull/140/checks?check_run_id=1913665717#step:6:55

Testing here to see if it is the build or not.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
       matrix:
         version:
           - '1.3'
+          - '1.6-nightly'
           - '1'
           - 'nightly'
         os:


Alexander-Barth:
`exit code: 3221225477`: yes windows error are as helpful as they have always been. 
Apparent, in hex it is 0xc0000005 should mean access violation in Windows.

visr:
Ah thanks. So it seems like an error in NetCDF.jl, which is good, so I'll close this.

Still funny that it is only triggered on Windows 1.6+. Nightly has the same issue, slighly better error: https://github.com/JuliaGeo/NetCDF.jl/pull/140/checks?check_run_id=1913665812#step:6:57

PR #128, opened 2021-04-19
mention Glob.jl

### LINKED ISSUE #127: Allow `*` syntax for multi file loading
Python's `xarray` has a nice feature when loading multifile datasets. You can do:

```julia
ERA5_files = "ERA5_monthly3D_reanalysis_*.nc"
xa = xr.open_mfdataset(ERA5_files)
```
and will match all files with the `*` getting all possible values the files have in the directory.

Would be nice if this happened here as well when input string contained the `*` character. 

Would this be accepted as a PR? 

Any tips on how to do this, e.g. using Regex?

Datseris:
Closes #127


Datseris:
Thanks @visr I didn't know of Glob.jl. This will helps others that also do not know about it.

Alexander-Barth:
Thanks @Datseris for updating the docs!

PR #131, opened 2021-04-28
docs: fix dim field name

visr:
https://github.com/Alexander-Barth/NCDatasets.jl/blob/5cb91db7e2926274e5e408d93d6fa352485c094a/src/dataset.jl#L51-L60


Alexander-Barth:
Thanks a lot @visr !

PR #136, opened 2021-08-06
Reduce compilation time

rafaqz:
Compilation takes a long time for NCDatasets on the first few function calls. This PR explores strategies to reduce the compile time for loading `NCDatasets` and `CFVariables`

- type stability of struct fields
- `precompile` functions
- `Dict{Symbol,Any}` instead of mixed type `NamedTuple`

Original:
```julia
julia>     using NCDatasets

julia>     filename = "/home/raf/.julia/dev/GeoData/test/data/tos_O1_2001-2002.nc"
"/home/raf/.julia/dev/GeoData/test/data/tos_O1_2001-2002.nc"

julia>     @time ds = NCDatasets.NCDataset(filename);
  0.827610 seconds (1.86 M allocations: 103.473 MiB, 8.01% gc time, 99.75% compilation time)

julia>     @time NCDatasets.variable(ds, "tos");
  0.000147 seconds (10 allocations: 880 bytes)

julia>     @time ds["tos"];
  0.570029 seconds (1.47 M allocations: 87.999 MiB, 26.39% gc time, 99.70% compilation time)
```
This PR:
```julia
julia>     using NCDatasets
[ Info: Precompiling NCDatasets [85f8d34a-cbdd-5861-8df4-14fed0d494ab]

julia>     filename = "/home/raf/.julia/dev/GeoData/test/data/tos_O1_2001-2002.nc"
"/home/raf/.julia/dev/GeoData/test/data/tos_O1_2001-2002.nc"

julia>     @time ds = NCDatasets.NCDataset(filename);
  0.626486 seconds (494.88 k allocations: 29.255 MiB, 12.36% gc time, 99.73% compilation time)

julia>     @time NCDatasets.variable(ds, "tos");
  0.000045 seconds (10 allocations: 880 bytes)

julia>     @time ds["tos"];
  0.134726 seconds (74.41 k allocations: 4.506 MiB, 99.38% compilation time)
```

It's especially faster (4x) for `getindex(::NCDataset, ::String)`. There are a lot of other optimizations that could be doe, but I'm opening this as is for now.

### DIFF
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -56,6 +56,7 @@ include("cfconventions.jl")
 include("defer.jl")
 include("multifile.jl")
 include("ncgen.jl")
+include("precompile.jl")
 
 
 end # module
--- a/src/cfvariable.jl
+++ b/src/cfvariable.jl
@@ -251,7 +251,7 @@ export defVar
 
 function _boundsParentVar(ds,name)
     # get from cache is available
-    if ds._boundsmap !== nothing
+    if length(values(ds._boundsmap)) === 0
         return get(ds._boundsmap,name,"")
     else
         for vname in keys(ds)
@@ -352,6 +352,7 @@ function Base.getindex(ds::AbstractDataset,varname::SymbolOrString)
         calendar,time_origin,time_factor = _calendar_time(v.attrib)
     else
         calendar,time_origin,time_factor = _calendar_time(variable(ds,parentname).attrib)
+        @show calendar
         calendar,time_origin,time_factor = _calendar_time(
             v.attrib;
             calendar = calendar,
@@ -373,14 +374,13 @@ function Base.getindex(ds::AbstractDataset,varname::SymbolOrString)
         end
     end
 
-
-    storage_attrib = Dict{Symbol,Any}(
-        :fillvalue => fillvalue,
-        :scale_factor => scale_factor,
-        :add_offset => add_offset,
-        :calendar => calendar,
-        :time_origin => time_origin,
-        :time_factor => time_factor,
+    storage_attrib = (
+        fillvalue = fillvalue,
+        scale_factor = scale_factor,
+        add_offset = add_offset,
+        calendar = calendar,
+        time_origin = time_origin,
+        time_factor = time_factor,
     )
 
     rettype = _get_rettype(ds, calendar, fillvalue, scaledtype)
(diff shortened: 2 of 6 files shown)


Alexander-Barth:
Thanks for taking a look into precompile latency. The idea to use a named tuple for `storage_attrib` was that the compiler could generate fully type stable and specialized code when applying the scaling (scale_factor and add_offset) and the masking.
With your PR, would e.g. `fillvalue(v)` not become type unstable?

rafaqz:
Yes there will be some Any fuelds. But its about a tradeoff... Up front compilatuon vs some type instability later, and only having to compile the expensive functions once. With the named tuple you have many different functions to compile, and getindex is much slower. Using a Dict seemed to help that a lot because there is nothing to infer.

But Dict/NamedTuple is probably not the most important thing here. The precompile helps a lot, and most structs in this package are currently not type stable, which adds a lot to compilation time and was relatively easy to fix for the main types. I didnt fix structs in e.g deffered.jl, but theyre also not stable.

rafaqz:
Not sure what the errors are, something breaking in older versions as I've only tested this on 1.6

Alexander-Barth:
Can you amend your PR with keeping `storage_attrib`as a named tuple ? Maybe we can use also the `precompile` approach by using the most common combination of `fillvalue/scale_factor/add_offset/...` (but this could be a separate PR).

rafaqz:
Sure it will be interesting to see the comparison separate from the other changes. I meant to do this earlier I've been a little busy lately.

rafaqz:
With precompilation it looks like the `NamedTuple` is actually slightly faster. 

```julia
julia>     @time ds["tos"];
  0.094708 seconds (13.40 k allocations: 912.854 KiB, 98.85% compilation time)
```

I also fixed the type stability of multifile datasets and a few other minor things.

rafaqz:
The test failures are some kind shared library errors on Julia 1.3, not a problem with this PR.

Alexander-Barth:
I ran a git bisect with julia 1.3.1, and it seems that the issue arises first on commit 2671e4aeca9aa77915399fceb5292a094bf0e497 .

Alexander-Barth:
Commenting these lines in `precompile.jl` makes 2671e4a  succeed in Julia 1.3.1:

        #precompile(getindex, (CF, Colon))                                                                                                                            
        #precompile(getindex, (CF, Int))                                                                                                                              
        #precompile(getindex, (CF, ntuple(Int, N)...))                                                                                                                
        #precompile(Array, (CF,))                                                                                                                                     
        #precompile(Array, (Var,))               

Maybe a bug in Julia 1.3 ? One approach could be to wrap `precompile.jl` in `if VERSION >= v"1.6" ... end`

rafaqz:
Maybe compilation RAM usage or something. `>= v1.6` sounds like a good solution to me.

Alexander-Barth:
Thanks a lot for your contribution! I just merged your PR.

rafaqz:
Thanks!

PR #141, opened 2021-08-17
Fix a typo and use CF standard names

petershintech:


petershintech:
While reviewing the test code, I made a couple of changes.

Alexander-Barth:
Thanks Peter for reviewing the code and fixing these issues!

PR #156, opened 2022-01-27
feat: allow `Symbol` names for attributes

### LINKED ISSUE #154: Allow symbols in attrib dicts?
It's not really a bug, but the current implementation does not allow for `Symbol` types in the attribute dictionaries, e.g.:

```julia
julia> var = OrderedDict(:long_name => "variable")
OrderedDict{Symbol, String} with 1 entry:
  :long_name => "variable"

julia> NCDataset("/tmp/test.nc", "c") do dataset
           defDim(dataset, "lon", 100)
           defDim(dataset, "lat", 110)
           defVar(dataset, "var", String, ("lon", "lat"), attrib = var)
       end
ERROR: MethodError: no method matching setindex!(::NCDatasets.Attributes{NCDataset{Nothing}}, ::String, ::Symbol)
Closest candidates are:
  setindex!(::NCDatasets.Attributes, ::Any, ::AbstractString) at ~/vcs/NCDatasets.jl/src/attributes.jl:72
Stacktrace:
 [1] defVar(ds::NCDataset{Nothing}, name::String, vtype::DataType, dimnames::Tuple{String, String}; kwargs::Base.Pairs{Symbol, OrderedDict{Symbol, String}, Tuple{Symbol}, NamedTuple{(:attrib,), Tuple{OrderedDict{Symbol, String}}}})
   @ NCDatasets ~/vcs/NCDatasets.jl/src/cfvariable.jl:144
 [2] (::var"#1#2")(dataset::NCDataset{Nothing})
   @ Main ./REPL[7]:4
 [3] NCDataset(::var"#1#2", ::String, ::Vararg{String}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ NCDatasets ~/vcs/NCDatasets.jl/src/dataset.jl:220
 [4] NCDataset(::Function, ::String, ::Vararg{String})
   @ NCDatasets ~/vcs/NCDatasets.jl/src/dataset.jl:218
 [5] top-level scope
   @ REPL[7]:1
```

It surprised me, since symbols are allowed in `OrderedDict`s. I wonder if the type constraint on `setindex` and any other affected methods could be changed to `Union{AbstractString, Symbol}` without negative effects? Alternatively, perhaps this case deserves a special error message, something like "Symbols are not allowed in attribute dictionaries"?

Low priority, I'm willing to take this on if you don't have time.

adigitoleo:
Closes #154.

Removed duplicated `haskey` method for attributes.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -16,7 +16,7 @@ function listAtt(ncid,varid)
 end
 
 
-function Base.get(a::BaseAttributes, name::AbstractString,default)
+function Base.get(a::BaseAttributes, name::SymbolOrString,default)
     if haskey(a,name)
         return a[name]
     else
@@ -26,7 +26,7 @@ end
 
 
 """
-    getindex(a::Attributes,name::AbstractString)
+    getindex(a::Attributes,name::SymbolOrString)
 
 Return the value of the attribute called `name` from the
 attribute list `a`. Generally the attributes are loaded by
@@ -37,13 +37,13 @@ ds = NCDataset("file.nc")
 title = ds.attrib["title"]
 ```
 """
-function Base.getindex(a::Attributes,name::AbstractString)
+function Base.getindex(a::Attributes,name::SymbolOrString)
     return nc_get_att(a.ds.ncid,a.varid,name)
 end
 
 
 """
-    Base.setindex!(a::Attributes,data,name::AbstractString)
+    Base.setindex!(a::Attributes,data,name::SymbolOrString)
 
 Set the attribute called `name` to the value `data` in the
 attribute list `a`. `data` can be a vector or a scalar. A scalar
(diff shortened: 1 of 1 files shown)


Alexander-Barth:
Thanks a lot for the PR!
The removed `haskey` is in fact an optimization (specialization) allowing to reduce the number of allocation introduced in response to:
https://github.com/CliMA/TurbulenceConvection.jl/issues/723 

Can you restore this in your PR? Can you also add a test?

Thanks again!

adigitoleo:
Ah, I removed it because it clashed with haskey from dataset.jl (L371). One possibility is to change `NCIterable` -> `Union{AbstrractDimensions,AbstractDataset,AbstractGroups}` (or introduce a new type alias for that). I'm not sure if that's a good idea though?

I'll work on a test tonight or tomorrow, thanks for the reminder.

Alexander-Barth:
I think I found a way to avoid the ambiguity. I will merge the PR (as I do not know if/how I can modify our PR ...)

Thank you for your contribution !

Issue #101, opened 2020-09-09
Feature request: indexing datasets with symbols?


ali-ramadhan:
I don't know much about how the NetCDF C library works but right now you must index the dimensions and variables of an `NCDataset` with a string, e.g. `ds["time"]` and I was wondering if it would be possible to allow indexing via symbols, e.g. `ds[:time]`?

Is it as easy as dispatching like
```julia
getindex(ds::AbstractDataset, sym::Symbol) = getindex(ds, string(sym))
```
?

cc @glwagner

visr:
Sounds good to me. Note that a ccall that expects a Cstring argument will happily accept either String or Symbol, so it should just be a matter of widening signatures to include symbols, no need for conversion.

Alexander-Barth:
I guess, string indexing needs to stay because it is easier to handle variable names with space (or operators) in them or composed of a prefix and suffix (but is also possible with Symbols). So we should support both ways, string and symbol indexing? Would the user code not start to look ugly/confusing if they start to mix string and symbol ? Or is it upon the user to be consistent  ?

visr:
Yeah AbstractString indexing definitely needs to stay, so we'd have to support both ways. I think the consistency here would indeed be up to the user. We could be consistent and always use string in the docs for instance.

This has precedence with [DataFrames.jl indexing](https://juliadata.github.io/DataFrames.jl/stable/lib/indexing/). Internally they use this in the type signatures:

```julia
const SymbolOrString = Union{Symbol, AbstractString}
```

Alexander-Barth:
Thanks @visr, `SymbolOrString` trick seems to work here as well. But I realize now we probably need to allow symbols also for attribute, group and dimension names too.

Alexander-Barth:
An alternative idea is to use symbols for query a NetCDF variable by standard name (CF convensions).

For instance if you have a variable with the poetic name `zos` in a NetCDF file and which has the following attributes:

```
zos (112 × 61 × 2)                                                                                                                                                   
  Datatype:    Float32                                                                                                                                                
  Dimensions:  lon × lat × time                                                                                                                                       
  Attributes:                                                                                                                                                         
   _FillValue           = 1.0e20                                                                                                                                      
   standard_name        = sea_surface_height_above_geoid                                                                                                              
   coordinates          = time lat lon                                                                                                                                
   long_name            = sea surface height                                                                                                                          
   units                = m                                                                                                                                           
[...]

PR #157, opened 2022-01-27
Symbol attributes

adigitoleo:
Fixes breakage due to #156. Added a test to catch this, confirmed that it fails without 246dedb.


adigitoleo:
As per discussion #101, we can just let `ccall` handle the Symbol/String type.

adigitoleo:
Hmm test is not quite right, hold on.

Alexander-Barth:
Thanks a lot checking this feature!

PR #159, opened 2022-02-03
feat: Add show method for dimensions

adigitoleo:


Alexander-Barth:
Thanks a lot for the PR!

PR #172, opened 2022-04-17
add capability to read only part of continuous ragged array

gaelforget:
This modification allows calling `loadragged` with either a range (e.g., `index=10:20`) or an integer. Currently the only option is `index=:` which can be limiting, notably for larger data sets.

### DIFF
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -118,7 +118,7 @@ end
 
 
 """
-     data = loadragged(ncvar,index::Colon)
+     data = loadragged(ncvar,index::Union{Colon,UnitRange,Int})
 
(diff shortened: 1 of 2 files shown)


Alexander-Barth:
This is great, thanks! Can you add also a test case, maybe here?
https://github.com/Alexander-Barth/NCDatasets.jl/blob/master/test/test_cont_ragged_array.jl

gaelforget:
> This is great, thanks! Can you add also a test case, maybe here?
> https://github.com/Alexander-Barth/NCDatasets.jl/blob/master/test/test_cont_ragged_array.jl

I just did in [2e8c426](https://github.com/Alexander-Barth/NCDatasets.jl/pull/172/commits/2e8c426db7107c3085fae0e6877d9bb23322fe8b). Doe this work?

Alexander-Barth:
Thank you very much!

Issue #150, opened 2021-11-30
Broader platform support


Alexander-Barth:
It would be great if NCDatasets.jl could be run on all platforms on which julia is [supported](https://julialang.org/downloads/#currently_supported_platforms). However, currently only on these platforms we have a NetCDF binary:

```
 NetCDF.v400.702.400.aarch64-linux-gnu.tar.gz
 NetCDF.v400.702.400.i686-w64-mingw32.tar.gz
 NetCDF.v400.702.400.x86_64-apple-darwin.tar.gz
 NetCDF.v400.702.400.x86_64-linux-gnu.tar.gz
 NetCDF.v400.702.400.x86_64-w64-mingw32.tar.gz
```
Installation of NCDatasets.jl will fail on other platforms.
Notably, this leaves out Raspberry PIs (32-bit armv7l), Linux on PowerPC, FreeBSD, Linux with musl,  Apple M1,...

To support more architectures the [NetCDF_jll package](https://github.com/JuliaPackaging/Yggdrasil/blob/master/N/NetCDF/) and [HDF5_jll package](https://github.com/JuliaPackaging/Yggdrasil/blob/master/H/HDF5/)  would need to be updated. As a first test, it would be useful to know if native compilation (of the NetCDF sources) on these platforms pass all the tests (`make check`) and then attempt cross-compilation (using BinaryBuilder).

ref https://github.com/Alexander-Barth/NCDatasets.jl/issues/116

visr:
Regarding cross-compilation, after tracing the breadcrumbs again it seems we now have an upstream issue we can refer to: https://github.com/HDFGroup/hdf5/issues/1203.

truedichotomy:
I'm very interested having NCDatasets.jl support Apple Silicon. I just installed netcdf 4.8.1 and HDF5 1.12.1 via homebrew on my M1 mac and they seem to function fine without having done too deep testing.

Alexander-Barth:
The first step would be to make a PR to update these build instructions to support Apple M1:

https://github.com/JuliaPackaging/Yggdrasil/blob/master/H/HDF5/build_tarballs.jl

truedichotomy:
I proposed some changes, but I don't know how to test whether they work.

Alexander-Barth:
One should be able to test it with this command (https://docs.binarybuilder.org/stable/):

```bash
julia --color=yes build_tarballs.jl --verbose
```

[...]

PR #193, opened 2022-12-08
Check if NetCDF_jll `is_available` before loading code

glwagner:
This change makes it possible to `using NCDatasets` on systems without NetCDF (due to eg #150).

X-ref: https://github.com/CliMA/Oceananigans.jl/pull/2847

cc @navidcy


glwagner:
We also need 12.10 to be registered to solve this problem over at Oceananigans after this PR is merged!

navidcy:
Nice! (Similar to https://github.com/JuliaGPU/AMGX.jl/pull/24)

Alexander-Barth:
Thanks for the PR! I will release a new version soon.

PR #195, opened 2022-12-09
Add support for `_normalizeindex` with `Base.OneTo`

charleskawczynski:
This PR adds support for `_normalizeindex` with `Base.OneTo` for the second argument. In ClimaCoreTempestRemap.jl, we use `axes` on a struct which returns a `Base.OneTo` type, and it's failing with `unsupported index`.

### DIFF
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -375,7 +375,7 @@ function Base.setindex!(v::Variable{T,N},data::AbstractArray{T2,N},indexes::Colo
     return data
 end
 
-
+_normalizeindex(n,ind::Base.OneTo) = 1:1:ind.stop
 _normalizeindex(n,ind::Colon) = 1:1:n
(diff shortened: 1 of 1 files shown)


charleskawczynski:
Hmm, I'm actually not sure whether this should be 
```julia
_normalizeindex(n, ind::Base.OneTo) = 1:1:ind.stop
```
or
```julia
_normalizeindex(n, ind::Base.OneTo) = 1:1:n
```

Alexander-Barth:
To me `ind.stop` seems correct, as `n` is the upper bound for the colon
type.

(Thanks for the pr!)

Charles Kawczynski ***@***.***> schrieb am Fr., 9. Dez. 2022,
19:30:

Alexander-Barth:
Thanks again!

PR #205, opened 2023-04-14
DiskArrays for `Variable`'s

tcarion:
This is an attempt to make the subtypes of `AbstractVariable` behave like `DiskArrays`. See JuliaGeo/CommonDataModel.jl#8.
I'm not perfectly sure of what I'm doing, but simple cases already work:

```julia
using NCDatasets
ds = NCDataset("tos_O1_2001-2002.nc")
var[:,:,1]
julia> var[:,:,1]
180×170 reshape(::Array{Union{Missing, Float32}, 3}, 180, 170) with eltype Union{Missing, Float32}:
 missing  missing  missing  missing  missing  missing  missing  missing  missing  …  271.419  271.42   271.422  271.426  271.43   271.437  271.445  271.459
 missing  missing  missing  missing  missing  missing  missing  missing  missing     271.419  271.42   271.422  271.426  271.431  271.438  271.445  271.459
 missing  missing  missing  missing  missing  missing  missing  missing  missing     271.418  271.42   271.422  271.425  271.431  271.438  271.445  271.459
 ⋮                                            ⋮                                   ⋱                               ⋮                                 
 missing  missing  missing  missing  missing  missing  missing  missing  missing     271.42   271.421  271.423  271.426  271.43   271.437  271.445  271.459
 missing  missing  missing  missing  missing  missing  missing  missing  missing     271.42   271.42   271.423  271.426  271.43   271.437  271.445  271.459
```

But of course it doesn't always work:
```julia
julia> var[2:4,1:3,1]
ERROR: BoundsError: attempt to access 3×3×1 Array{Union{Missing, Float32}, 3} at index [2:4, 1:3, 1:1]
Stacktrace:
  [1] throw_boundserror(A::Array{Union{Missing, Float32}, 3}, I::Tuple{UnitRange{Int64}, UnitRange{Int64}, UnitRange{Int64}})
    @ Base ./abstractarray.jl:691
  [2] checkbounds
    @ ./abstractarray.jl:656 [inlined]
  [3] view(::Array{Union{Missing, Float32}, 3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::UnitRange{Int64})
    @ Base ./subarray.jl:177
  [4] maybeview(::Array{Union{Missing, Float32}, 3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Vararg{UnitRange{Int64}})
    @ Base ./views.jl:146
  [5] dotview(::Array{Union{Missing, Float32}, 3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Vararg{UnitRange{Int64}})
    @ Base.Broadcast ./broadcast.jl:1200
  [6] readblock!(::NCDatasets.Variable{Float32, 3, NCDataset{Nothing}}, ::Array{Union{Missing, Float32}, 3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Vararg{UnitRange{Int64}})
    @ NCDatasets ~/.julia/dev/NCDatasets/src/variable.jl:449
  [7] readblock!(::CommonDataModel.CFVariable{Union{Missing, Float32}, 3, NCDatasets.Variable{Float32, 3, NCDataset{Nothing}}, NCDatasets.Attributes{NCDataset{Nothing}}, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float32, Tuple{Float32}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, ::Array{Union{Missing, Float32}, 3}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Vararg{UnitRange{Int64}})
    @ CommonDataModel ~/.julia/dev/CommonDataModel/src/cfvariable.jl:389
  [8] getindex_disk(::CommonDataModel.CFVariable{Union{Missing, Float32}, 3, NCDatasets.Variable{Float32, 3, NCDataset{Nothing}}, NCDatasets.Attributes{NCDataset{Nothing}}, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float32, Tuple{Float32}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, ::UnitRange{Int64}, ::Vararg{Any})
    @ DiskArrays ~/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:31
  [9] getindex(::CommonDataModel.CFVariable{Union{Missing, Float32}, 3, NCDatasets.Variable{Float32, 3, NCDataset{Nothing}}, NCDatasets.Attributes{NCDataset{Nothing}}, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float32, Tuple{Float32}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::Int64)
    @ DiskArrays ~/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:177
 [10] top-level scope
    @ REPL[17]:1
```

I will try to solve this in the coming days, any help is of course welcomed, and as I said above, I'm unsure about the approach I'm following :)

cc @rafaqz @Alexander-Barth

### DIFF
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -38,6 +38,7 @@ import CommonDataModel: AbstractDataset, AbstractVariable,
     groupnames, group,
     dimnames, dim,
     attribnames, attrib
+import DiskArrays: readblock!, writeblock!
 
 function __init__()
     NetCDF_jll.is_available() && init_certificate_authority()
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -39,6 +39,7 @@ import CommonDataModel: AbstractDataset, AbstractVariable,
     dimnames, dim,
     attribnames, attrib
 import DiskArrays: readblock!, writeblock!
+using DiskArrays: @implement_diskarray
 
 function __init__()
     NetCDF_jll.is_available() && init_certificate_authority()
@@ -65,6 +66,8 @@ include("multifile.jl")
 include("ncgen.jl")
 include("select.jl")
 include("precompile.jl")
+
+@implement_diskarray NCDatasets.Variable
 
 export CatArrays
 export CFTime
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -38,7 +38,8 @@ import CommonDataModel: AbstractDataset, AbstractVariable,
     groupnames, group,
     dimnames, dim,
     attribnames, attrib
-import DiskArrays: readblock!, writeblock!
+import DiskArrays
+import DiskArrays: readblock!, writeblock!, eachchunk, haschunks
 using DiskArrays: @implement_diskarray
 
 function __init__()
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -387,18 +387,6 @@ function _write_data_to_nc(v::Variable, data, indexes::Union{AbstractRange{<:Int
     return _write_data_to_nc(v, data, ind...)
 end
 
-function grow!(v::CFVariable, data, indexes::Union{Integer, Colon}...)
-    unlimdims = unlimited(Dimensions(CommonDataModel.dataset(v)))
-    length(unlimdims) == 1 || error("Only 1 unlimited dimension is supported") 
-    alldims = dimnames(v)
-    iunlimdim = findfirst(==(unlimdims[1]), alldims)
-    icol = findfirst(x -> x isa Colon, indexes)
-    iunlimdim == icol || ArgumentError("The Colon in the arguments must match the unlimited dimension.")
-    inds = replace(indexes, Colon() => 1:length(data))
-    v[inds...] = data
-    return v
-end
-
 getchunksize(v::Variable) = getchunksize(haschunks(v),v)
 getchunksize(::DiskArrays.Chunked, v::Variable) = chunking(v)[2]
 # getchunksize(::DiskArrays.Unchunked, v::Variable) = DiskArrays.estimate_chunksize(v)
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -368,7 +368,9 @@ getchunksize(v::Variable) = getchunksize(haschunks(v),v)
 getchunksize(::DiskArrays.Chunked, v::Variable) = chunking(v)[2]
 # getchunksize(::DiskArrays.Unchunked, v::Variable) = DiskArrays.estimate_chunksize(v)
 getchunksize(::DiskArrays.Unchunked, v::Variable) = size(v)
-eachchunk(v::Variable) = DiskArrays.GridChunks(v, getchunksize(v))
+eachchunk(v::CFVariable) = eachchunk(v.var)
+haschunks(v::CFVariable) = haschunks(v.var)
+eachchunk(v::Variable) = DiskArrays.GridChunks(v, Tuple(getchunksize(v)))
 haschunks(v::Variable) = (chunking(v)[1] == :contiguous ? DiskArrays.Unchunked() : DiskArrays.Chunked())
 
 _normalizeindex(n,ind::Base.OneTo) = 1:1:ind.stop
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -349,11 +349,6 @@ end
 
 _write_data_to_nc(v::Variable, data) = _write_data_to_nc(v, data, 1)
 
-function _write_data_to_nc(v::Variable, data, indexes::Colon...)
-    @info "Am I even getting here ?"
-    nc_put_var(v.ds.ncid,v.varid,data)
-end
-
 function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
     start,count,stride,jlshape = ncsub(indexes)
     nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -309,7 +309,7 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-function readblock!(v::Variable, aout, indexes::AbstractVector...)
+function readblock!(v::Variable, aout, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
     datamode(v.ds)
     _read_data_from_nc!(v, aout, indexes...)
     return aout
@@ -337,7 +337,7 @@ _read_data_from_nc!(v::Variable, aout) = _read_data_from_nc!(v, aout, 1)
 # readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
 # writeblock!(v::Variable{T, 0}, data) where T = writeblock!(v, data, 1)
 
-function writeblock!(v::Variable, data, indexes::AbstractVector...)
+function writeblock!(v::Variable, data, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
     datamode(v.ds)
     _write_data_to_nc(v, data, indexes...)
     return data
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -309,11 +309,21 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-function readblock!(v::Variable, aout, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
+# This method needs to be duplicated instead of using an Union. Otherwise a DiskArrays fallback is called instead which impacts performances 
+# (see https://github.com/Alexander-Barth/NCDatasets.jl/pull/205#issuecomment-1589575041)
+function readblock!(v::Variable, aout, indexes::AbstractUnitRange...)
     datamode(v.ds)
     _read_data_from_nc!(v, aout, indexes...)
     return aout
 end
+
+function readblock!(v::Variable, aout, indexes::StepRange...)
+    datamode(v.ds)
+    _read_data_from_nc!(v, aout, indexes...)
+    return aout
+end
+
+readblock!(v::Variable, aout) = _read_data_from_nc!(v::Variable, aout)
 
 function _read_data_from_nc!(v::Variable, aout, indexes::Int...)
     aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in reverse(indexes)])
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -508,7 +508,40 @@ function writeblock!(v::Variable,data,indexes::Union{Int,Colon,AbstractRange{<:I
 end
 
 
-readblock!(v::Union{MFVariable,DeferVariable,Variable}, aout, ci::CartesianIndices) = aout .= v[ci.indices...]
-writeblock!(v::Union{MFVariable,DeferVariable,Variable},data,ci::CartesianIndices) = writeblock!(v,data,ci.indices...)
+readblock!(v::Variable, aout, ci::CartesianIndices) = aout .= v[ci.indices...]
+writeblock!(v::Variable, data, ci::CartesianIndices) = writeblock!(v,data,ci.indices...)
 
+Base.getindex(v::Union{MFVariable,DeferVariable},ci::CartesianIndices) = v[ci.indices...]
+Base.setindex!(v::Union{MFVariable,DeferVariable},data,ci::CartesianIndices) = setindex!(v,data,ci.indices...)
 
+function readblock!(v::Variable, aout, indices::Union{Int,Colon,AbstractRange{<:Integer},Vector{Int}}...)
+    @debug "transform vector of indices to ranges"
+
+    sz_source = size(v)
+    ri = to_range_list.(indices,sz_source)
+    sz_dest = NCDatasets._shape_after_slice(sz_source,indices...)
+
+    N = length(indices)
+
+    ri_dest = range_indices_dest(ri...)
+    @debug "ri_dest $ri_dest"
+    @debug "ri $ri"
+
+    if all(==(1),length.(ri))
+        # single chunk
+        R = first(CartesianIndices(length.(ri)))
+        ind_source = ntuple(i -> ri[i][R[i]],N)
+        ind_dest = ntuple(i -> ri_dest[i][R[i]],length(ri_dest))
+        return aout[indices...] .= v[ind_source...]
+    end
+
+    dest = Array{eltype(v),length(sz_dest)}(undef,sz_dest)
+    for R in CartesianIndices(length.(ri))
+        ind_source = ntuple(i -> ri[i][R[i]],N)
+        ind_dest = ntuple(i -> ri_dest[i][R[i]],length(ri_dest))
+        #dest[ind_dest...] = v[ind_source...]
+        buffer = Array{eltype(v.var),length(ind_dest)}(undef,length.(ind_dest))
+        load!(v,view(dest,ind_dest...),buffer,ind_source...)
+    end
+    return aout[indices...] .= dest
+end
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -341,7 +341,6 @@ readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
 # writeblock!(v::Variable{T, 0}, data) where T = writeblock!(v, data, 1)
 
 function writeblock!(v::Variable{T,N},data,indexes::AbstractUnitRange...) where {T,N}
-    @debug "$(@__LINE__) wb $((data, indexes))"
     datamode(v.ds)
     _write_data_to_nc(v, data, indexes...)
     return data
@@ -351,19 +350,20 @@ function _write_data_to_nc(v::Variable{T,N},data,indexes::Int...) where {T,N}
     nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]],T(data[1]))
 end
 
-_write_data_to_nc(v::Variable,data) = _write_data_to_nc(v, data, 1)
+_write_data_to_nc(v::Variable, data) = _write_data_to_nc(v, data, 1)
 
-function _write_data_to_nc(v::Variable,data,indexes::Colon...)
+function _write_data_to_nc(v::Variable, data, indexes::Colon...)
+    @info "Am I even getting here ?"
     nc_put_var(v.ds.ncid,v.varid,data)
 end
 
-function _write_data_to_nc(v::Variable,data,indexes::StepRange{Int,Int}...)
+function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
     start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
-    nc_put_vars(v.ds.ncid,v.varid,start,count,stride,data)
+    nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
 end
 
-function _write_data_to_nc(v::Variable,data,indexes::Union{Int,Colon,AbstractRange{<:Integer}}...)
-    ind = normalizeindexes(size(v),indexes)
+function _write_data_to_nc(v::Variable, data, indexes::AbstractRange{<:Integer}...)
+    ind = prod(length.(indexes)) == 1 ? first.(indexes) : normalizeindexes(size(v),indexes)
     return _write_data_to_nc(v, data, ind...)
 end
 
@@ -419,9 +419,6 @@ end
 
     return start,count,stride
 end
-
-readblock!(v::Variable, aout, ci::CartesianIndices) = aout .= v[ci.indices...]
-writeblock!(v::Variable, data, ci::CartesianIndices) = writeblock!(v,data,ci.indices...)
 
 Base.getindex(v::Union{MFVariable,DeferVariable},ci::CartesianIndices) = v[ci.indices...]
 Base.setindex!(v::Union{MFVariable,DeferVariable},data,ci::CartesianIndices) = setindex!(v,data,ci.indices...)
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -309,35 +309,32 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-
-function readblock!(v::Variable, aout, indexes::Int...)
+function readblock!(v::Variable, aout, indexes::AbstractUnitRange...)
     datamode(v.ds)
-    aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
-end
-
-function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
-    start,count,stride,jlshape = ncsub(indexes[1:N])
-    data = Array{T,N}(undef,jlshape)
-    datamode(v.ds)
-    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
-    # nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
-    aout .= data
+    _read_data_from_nc!(v, aout, indexes...)
     return aout
 end
 
-function readblock!(v::Variable{T,N}, aout, indexes::Union{Int,Colon,AbstractRange{<:Integer}}...) where {T,N}
+function _read_data_from_nc!(v::Variable, aout, indexes::Int...)
+    aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
+end
+
+function _read_data_from_nc!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
+    start,count,stride,jlshape = ncsub(indexes[1:N])
+    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
+end
+
+function _read_data_from_nc!(v::Variable{T,N}, aout, indexes::Union{Int,Colon,AbstractRange{<:Integer}}...) where {T,N}
     sz = size(v)
     start,count,stride = ncsub2(sz,indexes...)
     jlshape = _shape_after_slice(sz,indexes...)
-    data = Array{T}(undef,jlshape)
-
-    datamode(v.ds)
-    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
-    aout .= data
+    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
 end
 
+_read_data_from_nc!(v::Variable, aout) = _read_data_from_nc!(v, aout, 1)
+
 # NetCDF scalars indexed as []
-readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
+# readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
 # writeblock!(v::Variable{T, 0}, data) where T = writeblock!(v, data, 1)
 
 function writeblock!(v::Variable{T,N},data,indexes::AbstractUnitRange...) where {T,N}
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -443,12 +443,12 @@ end
 
 function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
     start,count,stride,jlshape = ncsub(indexes[1:N])
-    data = Array{T,N}(undef,jlshape)
-
+    # data = Array{T,N}(undef,jlshape)
     datamode(v.ds)
-    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
-    aout .= data
-    return data
+    # nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
+    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
+    # aout .= data
+    return aout
 end
 
 function writeblock!(v::Variable{T,N},data::T,indexes::StepRange{Int,Int}...) where {T,N}
@@ -488,7 +488,7 @@ function readblock!(v::Variable{T,N}, aout, indexes::Union{Int,Colon,AbstractRan
 
     datamode(v.ds)
     nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
-    aout[indexes...] .= data
+    aout .= data
 end
 
 # NetCDF scalars indexed as []
@@ -508,7 +508,7 @@ function writeblock!(v::Variable,data,indexes::Union{Int,Colon,AbstractRange{<:I
 end
 
 
-readblock!(v::Union{MFVariable,DeferVariable,Variable}, aout, ci::CartesianIndices) = aout[ci.indices...] .= v[ci.indices...]
+readblock!(v::Union{MFVariable,DeferVariable,Variable}, aout, ci::CartesianIndices) = aout .= v[ci.indices...]
 writeblock!(v::Union{MFVariable,DeferVariable,Variable},data,ci::CartesianIndices) = writeblock!(v,data,ci.indices...)
 
 
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -446,7 +446,8 @@ function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR
     data = Array{T,N}(undef,jlshape)
 
     datamode(v.ds)
-    aout[indexes...] .= nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
+    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
+    aout .= data
     return data
 end
 
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -312,7 +312,7 @@ export nomissing
 
 function readblock!(v::Variable, aout, indexes::Int...)
     datamode(v.ds)
-    aout[indexes...] .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
+    aout[indexes...] = nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
 end
 
 function writeblock!(v::Variable{T,N},data,indexes::Int...) where N where T
@@ -356,7 +356,7 @@ for data_type = [Number, String, Char]
             return data
         end
 
-        writeblock!(v::Variable,data::$data_type,indexes::Colon...) = setindex!(v::Variable,data)
+        writeblock!(v::Variable,data::$data_type,indexes::Colon...) = writeblock!(v::Variable,data)
 
         function writeblock!(v::Variable{T,N},data::$data_type,indexes::StepRange{Int,Int}...) where {T,N}
             datamode(v.ds) # make sure that the file is in data mode
@@ -492,7 +492,8 @@ function readblock!(v::Variable{T,N}, aout, indexes::Union{Int,Colon,AbstractRan
 end
 
 # NetCDF scalars indexed as []
-readblock!(v::Variable{T, 0}, aout) where T = aout[1] = v[1]
+readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
+writeblock!(v::Variable{T, 0}, data::Array{T, 0}) where T = writeblock!(v, data[1])
 
 
 
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -333,10 +333,6 @@ end
 
 _read_data_from_nc!(v::Variable, aout) = _read_data_from_nc!(v, aout, 1)
 
-# NetCDF scalars indexed as []
-# readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
-# writeblock!(v::Variable{T, 0}, data) where T = writeblock!(v, data, 1)
-
 function writeblock!(v::Variable, data, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
     datamode(v.ds)
     _write_data_to_nc(v, data, indexes...)
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -359,6 +359,18 @@ function _write_data_to_nc(v::Variable, data, indexes::Union{AbstractRange{<:Int
     return _write_data_to_nc(v, data, ind...)
 end
 
+function grow!(v::CFVariable, data, indexes::Union{Integer, Colon}...)
+    unlimdims = unlimited(Dimensions(CommonDataModel.dataset(v)))
+    length(unlimdims) == 1 || error("Only 1 unlimited dimension is supported") 
+    alldims = dimnames(v)
+    iunlimdim = findfirst(==(unlimdims[1]), alldims)
+    icol = findfirst(x -> x isa Colon, indexes)
+    iunlimdim == icol || ArgumentError("The Colon in the arguments must match the unlimited dimension.")
+    inds = replace(indexes, Colon() => 1:length(data))
+    v[inds...] = data
+    return v
+end
+
 getchunksize(v::Variable) = getchunksize(haschunks(v),v)
 getchunksize(::DiskArrays.Chunked, v::Variable) = chunking(v)[2]
 # getchunksize(::DiskArrays.Unchunked, v::Variable) = DiskArrays.estimate_chunksize(v)
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -316,11 +316,11 @@ function readblock!(v::Variable, aout, indexes::AbstractUnitRange...)
 end
 
 function _read_data_from_nc!(v::Variable, aout, indexes::Int...)
-    aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
+    aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in reverse(indexes)])
 end
 
 function _read_data_from_nc!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
-    start,count,stride,jlshape = ncsub(indexes[1:N])
+    start,count,stride,jlshape = ncsub(indexes)
     nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
 end
 
@@ -344,7 +344,7 @@ function writeblock!(v::Variable{T,N},data,indexes::AbstractUnitRange...) where 
 end
 
 function _write_data_to_nc(v::Variable{T,N},data,indexes::Int...) where {T,N}
-    nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]],T(data[1]))
+    nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in reverse(indexes)],T(data[1]))
 end
 
 _write_data_to_nc(v::Variable, data) = _write_data_to_nc(v, data, 1)
@@ -355,7 +355,7 @@ function _write_data_to_nc(v::Variable, data, indexes::Colon...)
 end
 
 function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
-    start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
+    start,count,stride,jlshape = ncsub(indexes)
     nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
 end
 
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -309,7 +309,7 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-function readblock!(v::Variable, aout, indexes::AbstractUnitRange...)
+function readblock!(v::Variable, aout, indexes::AbstractVector...)
     datamode(v.ds)
     _read_data_from_nc!(v, aout, indexes...)
     return aout
@@ -337,7 +337,7 @@ _read_data_from_nc!(v::Variable, aout) = _read_data_from_nc!(v, aout, 1)
 # readblock!(v::Variable{T, 0}, aout) where T = readblock!(v, aout, 1)
 # writeblock!(v::Variable{T, 0}, data) where T = writeblock!(v, data, 1)
 
-function writeblock!(v::Variable{T,N},data,indexes::AbstractUnitRange...) where {T,N}
+function writeblock!(v::Variable, data, indexes::AbstractVector...)
     datamode(v.ds)
     _write_data_to_nc(v, data, indexes...)
     return data
@@ -359,7 +359,7 @@ function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}.
     nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
 end
 
-function _write_data_to_nc(v::Variable, data, indexes::AbstractRange{<:Integer}...)
+function _write_data_to_nc(v::Variable, data, indexes::Union{AbstractRange{<:Integer}}...)
     ind = prod(length.(indexes)) == 1 ? first.(indexes) : normalizeindexes(size(v),indexes)
     return _write_data_to_nc(v, data, ind...)
 end
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -310,12 +310,12 @@ nomissing(a::AbstractArray,value) = a
 export nomissing
 
 
-function Base.getindex(v::Variable,indexes::Int...)
+function readblock!(v::Variable, aout, indexes::Int...)
     datamode(v.ds)
-    return nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
+    aout[indexes...] .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
 end
 
-function Base.setindex!(v::Variable{T,N},data,indexes::Int...) where N where T
+function writeblock!(v::Variable{T,N},data,indexes::Int...) where N where T
     @debug "$(@__LINE__)"
     datamode(v.ds)
     # use zero-based indexes and reversed order
@@ -323,20 +323,20 @@ function Base.setindex!(v::Variable{T,N},data,indexes::Int...) where N where T
     return data
 end
 
-function Base.getindex(v::Variable{T,N},indexes::Colon...) where {T,N}
+function readblock!(v::Variable{T,N}, aout, indexes::Colon...) where {T,N}
     datamode(v.ds)
     data = Array{T,N}(undef,size(v))
     nc_get_var!(v.ds.ncid,v.varid,data)
 
     # special case for scalar NetCDF variable
     if N == 0
-        return data[]
+        aout[indexes...] .= data[]
     else
-        return data
+        aout[indexes...] .= data
     end
 end
 
-function Base.setindex!(v::Variable{T,N},data::T,indexes::Colon...) where {T,N}
+function writeblock!(v::Variable{T,N},data::T,indexes::Colon...) where {T,N}
     @debug "setindex! colon $data"
     datamode(v.ds) # make sure that the file is in data mode
     tmp = fill(data,size(v))
@@ -348,7 +348,7 @@ end
 for data_type = [Number, String, Char]
     @eval begin
         # call to v .= 123
-        function Base.setindex!(v::Variable{T,N},data::$data_type) where {T,N}
+        function writeblock!(v::Variable{T,N},data::$data_type) where {T,N}
             @debug "setindex! $data"
             datamode(v.ds) # make sure that the file is in data mode
             tmp = fill(convert(T,data),size(v))
@@ -356,9 +356,9 @@ for data_type = [Number, String, Char]
             return data
         end
 
-        Base.setindex!(v::Variable,data::$data_type,indexes::Colon...) = setindex!(v::Variable,data)
+        writeblock!(v::Variable,data::$data_type,indexes::Colon...) = setindex!(v::Variable,data)
 
-        function Base.setindex!(v::Variable{T,N},data::$data_type,indexes::StepRange{Int,Int}...) where {T,N}
+        function writeblock!(v::Variable{T,N},data::$data_type,indexes::StepRange{Int,Int}...) where {T,N}
             datamode(v.ds) # make sure that the file is in data mode
             start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
             tmp = fill(convert(T,data),jlshape)
@@ -368,14 +368,14 @@ for data_type = [Number, String, Char]
     end
 end
 
-function Base.setindex!(v::Variable{T,N},data::AbstractArray{T,N},indexes::Colon...) where {T,N}
+function writeblock!(v::Variable{T,N},data::AbstractArray{T,N},indexes::Colon...) where {T,N}
     datamode(v.ds) # make sure that the file is in data mode
 
     nc_put_var(v.ds.ncid,v.varid,data)
     return data
 end
 
-function Base.setindex!(v::Variable{T,N},data::AbstractArray{T2,N},indexes::Colon...) where {T,T2,N}
+function writeblock!(v::Variable{T,N},data::AbstractArray{T2,N},indexes::Colon...) where {T,T2,N}
     datamode(v.ds) # make sure that the file is in data mode
     tmp =
         if T <: Integer
@@ -441,16 +441,16 @@ end
     return start,count,stride
 end
 
-function Base.getindex(v::Variable{T,N},indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
+function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
     start,count,stride,jlshape = ncsub(indexes[1:N])
     data = Array{T,N}(undef,jlshape)
 
     datamode(v.ds)
-    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
+    aout[indexes...] .= nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
     return data
 end
 
-function Base.setindex!(v::Variable{T,N},data::T,indexes::StepRange{Int,Int}...) where {T,N}
+function writeblock!(v::Variable{T,N},data::T,indexes::StepRange{Int,Int}...) where {T,N}
     datamode(v.ds) # make sure that the file is in data mode
     start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
     tmp = fill(data,jlshape)
@@ -458,7 +458,7 @@ function Base.setindex!(v::Variable{T,N},data::T,indexes::StepRange{Int,Int}...)
     return data
 end
 
-function Base.setindex!(v::Variable{T,N},data::Array{T,N},indexes::StepRange{Int,Int}...) where {T,N}
+function writeblock!(v::Variable{T,N},data::Array{T,N},indexes::StepRange{Int,Int}...) where {T,N}
     datamode(v.ds) # make sure that the file is in data mode
     start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
     nc_put_vars(v.ds.ncid,v.varid,start,count,stride,data)
@@ -466,7 +466,7 @@ function Base.setindex!(v::Variable{T,N},data::Array{T,N},indexes::StepRange{Int
 end
 
 # data can be Array{T2,N} or BitArray{N}
-function Base.setindex!(v::Variable{T,N},data::AbstractArray,indexes::StepRange{Int,Int}...) where {T,N}
+function writeblock!(v::Variable{T,N},data::AbstractArray,indexes::StepRange{Int,Int}...) where {T,N}
     datamode(v.ds) # make sure that the file is in data mode
     start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
 
@@ -479,7 +479,7 @@ end
 
 
 
-function Base.getindex(v::Variable{T,N},indexes::Union{Int,Colon,AbstractRange{<:Integer}}...) where {T,N}
+function readblock!(v::Variable{T,N}, aout, indexes::Union{Int,Colon,AbstractRange{<:Integer}}...) where {T,N}
     sz = size(v)
     start,count,stride = ncsub2(sz,indexes...)
     jlshape = _shape_after_slice(sz,indexes...)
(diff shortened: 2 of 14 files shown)


rafaqz:
Hmm I dont get how your second example could ever work?

Alexander-Barth:
I have an old branch https://github.com/Alexander-Barth/NCDatasets.jl/tree/DiskArrays that might be useful.

What do you think about these changes?
```diff
diff --git a/src/NCDatasets.jl b/src/NCDatasets.jl
index 2140f5e..bd1da5a 100644
--- a/src/NCDatasets.jl
+++ b/src/NCDatasets.jl
@@ -38,7 +38,7 @@ import CommonDataModel: AbstractDataset, AbstractVariable,
     groupnames, group,
     dimnames, dim,
     attribnames, attrib
-import DiskArrays: readblock!, writeblock!
+import DiskArrays: readblock!, writeblock!, AbstractDiskArray
 
 function __init__()
     NetCDF_jll.is_available() && init_certificate_authority()
diff --git a/src/types.jl b/src/types.jl
index c89fb29..a637ce2 100644
--- a/src/types.jl
+++ b/src/types.jl
@@ -61,7 +61,7 @@ end
 
 # Variable (as stored in NetCDF file, without using
 # add_offset, scale_factor and _FillValue)
-mutable struct Variable{NetCDFType,N,TDS<:AbstractNCDataset} <: AbstractNCVariable{NetCDFType, N}
+mutable struct Variable{NetCDFType,N,TDS<:AbstractNCDataset} <: AbstractDiskArray{NetCDFType, N}
     ds::TDS
     varid::Cint
     dimids::NTuple{N,Cint}
diff --git a/src/variable.jl b/src/variable.jl
index 425df70..ba02c72 100644
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -446,8 +446,8 @@ function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR
     data = Array{T,N}(undef,jlshape)
 
     datamode(v.ds)
-    aout[indexes...] .= nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,data)
-    return data
+    nc_get_vars!(v.ds.ncid,v.varid,start,count,stride,aout)
+    return aout
 end
 
 function writeblock!(v::Variable{T,N},data::T,indexes::StepRange{Int,Int}...) where {T,N}
```

Alexander-Barth:
Unfortunately, we still have an infinite recursion when loading a scalar:

https://github.com/Alexander-Barth/NCDatasets.jl/blob/pr-DiskArray/test/test_scalar.jl#L16

rafaqz:
Indexing a muktidimensional array with a Colon like that is again breaking the base julia the array interface ;)

Alexander-Barth:
the data in question is a zero-dimensional array. This does work in base julia:

```julia
zero_dim_array =  dropdims([1],dims=1)
zero_dim_array[:]
```

`[:]` just flattens the array:

```
julia> ones(2,2)[:]
4-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
```

For some reasons my commits do not show in this PR, strange.

rafaqz:
Yes it should work, returning a vector.

But your testing that the result is a scalar?

rafaqz:
If you want to add diskarrays behavioir just to variables, you can do that using the macros provided in DiskArrays.jl, rather than suntyping AbstractDiskArray.

But Im not sure you would want any disk based array types to not have disk arrays behaviour, lazy chunked broadcasting is a pretty nice thing to have, instead if broadcasts just not working at all.

Alexander-Barth:
I am wondering if (at least as a start) we just convert the base type NCDatasets.Variable to derive from DiskArray. 
In this case, the conversion from Array{Union{Missing,T},N} to Array{T,N} is already taken care of.

Maybe the macros that @rafaqz mentioned can then be used for wrapped types like CFVariable without requiring that they derive from DiskArray?

You guys have a better understand of DiskArrays works... Is there any documentation about the macros? 

> Strange indeed! Do you commit to my forked diskarrays branch? Maybe I have to setup something in my forked repo?

I used the commands from https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally to test the PR:

```
git fetch origin pull/205/head:pr-DiskArray
git checkout pr-DiskArray 
```

and I assumed that any commits with show up here, but maybe that work only for PR that I initiate. In any case, if things become simplier for you, if you are a dev on this repo just let me know.

@rafaqz In general, yes some the test values should change. The issue of the recursion happened before but maybe it is related to this:

https://github.com/Alexander-Barth/NCDatasets.jl/pull/205/files#diff-fe41cd4463f1540b047828f14c99e312c078e817112d8f201a49eace1f5d58b9R495

I will already change the test in the master version.

tcarion:
I tried to run the test and I spotted a few general issues:

- When scalar variables are implied
- When infinite dimensions are implied
- When writing to the `Variable` with a different eltype ([e.g. here](https://github.com/Alexander-Barth/NCDatasets.jl/blob/690ba2206f5cd7f5395f9f92ad3d071ead950350/test/test_scaling.jl#L33))
- When dealing with other subtypes of `CommonDataModel.AbstractVariable` (`SubVariable`, `MFVariable`). That would imply to change all the `getindex` / `setindex!` methods with their `readblock!`/`writeblock` equivalent. That's why it might be more simple to add diskarrays behaviour only to Variable as you suggested above, but I don't know how to do this.

rafaqz:
See here for disk arrays macros. Above that application you can see all the sub macros. But you probably want the main one;

https://github.com/meggart/DiskArrays.jl/blob/main/src/DiskArrays.jl#L37



We could run your tests in CI so we can all see the failures and discuss them? I think thats the best way for me to contribute to the discussion.

@tcarion I will make you a JuliaGeo member so you dont need run permissions. You already have a package here anyway.

Alexander-Barth:
Concerning the error type "2. When infinite dimensions are implied" there are some changes API, for example when previously we could do 

```
# ncvar is a NetCDF vector variable with an ulimited dimension
ncvar[:] = data
```
we now need to do:

```
# ncvar is a NetCDF vector variable with an ulimited dimension
ncvar[1:length(data)] = data
```

which mimics the behaviour of NetCDF.jl.

For the other cases, however, I don't think (or hope :-) ) that we need to break current API.

Another point is that `ncvar[:]` currently loads all the data and preserving the shape (without flatting the array as in julia). This behavior can go away too. I tried to remove tests that rely on this behaviour, but I might have forgotten some.

Thanks @rafaqz for the info!

> So @implement_diskarray Variable should do it.

To be certain that I understand: we can either derive Variable from AbstractDiskArrays or we can use the macro to inherit the behavior (chunking,...) . Is this correct? Or do we need to do both?

What do you think about the approach to make just NCDatasets.Variable as a DiskArrays and all types wrapping NCDatasets.Variable would use method forwarding as [explained here](https://discourse.julialang.org/t/method-forwarding-macro/23355/4)?


I changed the setting in github action; I hope it works now. (If not maybe @tcarion can make a trivial change the to README.md so that he is not considered as a first time contributor to NCDatasets for github.)

tcarion:
> So @implement_diskarray Variable should do it.

I tried and I get a `UndefVarError: Variable not defined`. I think the problem is that DiskArrays looks for a `Variable` defined in its own Module:
```julia
julia> @macroexpand(@implement_getindex(Variable))
quote
    #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:177 =#
    (DiskArrays.Base).getindex(var"#574#a"::DiskArrays.Variable, var"#575#i"...) = begin
            #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:177 =#
            DiskArrays.getindex_disk(var"#574#a", var"#575#i"...)
        end
    #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:179 =#
    function (DiskArrays.Base).getindex(var"#578#a"::DiskArrays.Variable, var"#579#i"::DiskArrays.ChunkIndex)
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:179 =#
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:180 =#
        var"#576#cs" = DiskArrays.eachchunk(var"#578#a")
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:181 =#
        var"#577#inds" = var"#576#cs"[(var"#579#i").I]
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:182 =#
        return DiskArrays.wrapchunk((var"#579#i").chunktype, var"#578#a"[var"#577#inds"...], var"#577#inds")
    end
    #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:184 =#
    function (DiskArrays.DiskArrays).ChunkIndices(var"#580#a"::DiskArrays.Variable; offset = false)
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:184 =#
        #= /home/tcarion/.julia/packages/DiskArrays/zaCXL/src/diskarray.jl:185 =#
        return DiskArrays.ChunkIndices((DiskArrays.Base).OneTo.(DiskArrays.size(DiskArrays.eachchunk(var"#580#a"))), if offset
                    DiskArrays.OffsetChunks()
                else
                    DiskArrays.OneBasedChunks()
                end)
    end
end
```

I'm pretty unexperienced with macros, but I'd say it's something that should be fixed in DiskArrays, right?

> we now need to do:
> ```julia
> # ncvar is a NetCDF vector variable with an ulimited dimension
> ncvar[1:length(data)] = data
> ```
> which mimics the behaviour of NetCDF.jl.

Ok I'll keep that in mind when trying to make the tests work, thanks!


> What do you think about the approach to make just NCDatasets.Variable as a DiskArrays and all types wrapping NCDatasets.Variable would use method forwarding (for eachchunk...) as [explained here](https://discourse.julialang.org/t/method-forwarding-macro/23355/4)?

Since @implement_diskarray is currently not working, I will go with that approach for now.

> I changed the setting in github action; I hope it works now. (If not maybe @tcarion can make a trivial change the to README.md so that he is not considered as a first time contributor to NCDatasets for github.)

Thank you, I'll try after my next commit, I'll let you know if it didn't work :)

rafaqz:
@Alexander-Barth  `collect(ncvar)` or `Array(ncvar)` would be more Julian ways to do `ncvar[:]` in a generic way.

> we can either derive Variable from AbstractDiskArrays or we can use the macro to inherit the behavior (chunking,...) . Is this correct? Or do we need to do both?

Yes the macro is what we apply to `AbstractDiskArray` anyway, so it will give identical behavior to whatever you apply it to without inheriting from anything. You can also apply subsets of it with the other macros (see the lines above [here](https://github.com/meggart/DiskArrays.jl/blob/main/src/DiskArrays.jl#L37)

> What do you think about the approach to make just NCDatasets.Variable as a DiskArrays and all types wrapping NCDatasets.Variable would use method forwarding (for eachchunk...) as [explained here](https://discourse.julialang.org/t/method-forwarding-macro/23355/4)?

Honestly I think eventually every chunked disk-based array object in the julia ecosystem should use DiskArrays.jl so we have a consistent standard.

But we should of course compromise on that to just get things working! that's why we have the macros.

rafaqz:
@tcarion yeah this macro hasn't been very widely used, it seems to not be escaped properly.

Try using `CommonDataModel.Variable`

rafaqz:
Im workong on a fix now

rafaqz:
Turns oit there a few other fixes needed for this to work.

Alexander-Barth:
> @Alexander-Barth collect(ncvar) or Array(ncvar) would be more Julian ways to do ncvar[:] in a generic way.

I agree, we have already `Array(ncvar)` in NCDatasets. `collect` currently works too (but slow).

rafaqz:
I've registered a patch version of DiskArrays.jl with fixed macros, should be  hopefully that gets things working.

https://github.com/JuliaRegistries/General/pull/82137

tcarion:
Thank you very much @rafaqz, the macro works now!

Alexander-Barth:
@tcarion yes, the test can indeed be removed/changed. 

I make this change here: 0f0f8a9

Can you merge this change?
Thanks a lot for all your hard work on this issue !!

rafaqz:
As for the scalar indexing, have you tried it with another DiskArray? maybe try it with the`_DiskArray` object in the DiskArrays.jl tests or an ArchGDAL.jl `RasterDataset`

If its a DiskArrays.jl problem make an issue there and I'll fix it.

tcarion:
It wasn't a problem with DiskArrays :-) I could fix the issue with some trial and error (meaning that I don't fully understand what the problem was... but it works at least)

rafaqz:
```julia
function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
    start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
    nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
end
```

This allows arbitrary trailing ranges to pass? Shouldn't it error if there are the wrong number of range `indexes`?

Same with `Colon` - and extra colon normally adds a dimension.

Usually the only allowed trailing index for julia arrays is `1`.

src/variable.jl
```diff
@@ -351,19 +350,20 @@ function _write_data_to_nc(v::Variable{T,N},data,indexes::Int...) where {T,N}
     nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]],T(data[1]))
 end
 
-_write_data_to_nc(v::Variable,data) = _write_data_to_nc(v, data, 1)
+_write_data_to_nc(v::Variable, data) = _write_data_to_nc(v, data, 1)
 
-function _write_data_to_nc(v::Variable,data,indexes::Colon...)
+function _write_data_to_nc(v::Variable, data, indexes::Colon...)
+    @info "Am I even getting here ?"
```
rafaqz, line 356: I don't think `writeblock` will pass you a colon, just ranges.

src/variable.jl
```diff
@@ -351,19 +350,20 @@ function _write_data_to_nc(v::Variable{T,N},data,indexes::Int...) where {T,N}
...
     nc_put_var(v.ds.ncid,v.varid,data)
 end
 
-function _write_data_to_nc(v::Variable,data,indexes::StepRange{Int,Int}...)
+function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
     start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
-    nc_put_vars(v.ds.ncid,v.varid,start,count,stride,data)
+    nc_put_vars(v.ds.ncid,v.varid,start,count,stride,T.(data))
```
rafaqz, line 362: Probably DiskArrays.jl wont let you do this optimisation either. You may need to override these methods on `OrdinalRange` https://github.com/meggart/DiskArrays.jl/blob/main/src/batchgetindex.jl#L166-L216 so that can happen here rather than in DiskArrays.jl

tcarion:
> This allows arbitrary trailing ranges to pass? Shouldn't it error if there are the wrong number of range indexes?

I don't understand where the problem is? DiskArrays takes care of erroring when appropriate:
```julia
julia> v
var-Char (4 × 5)
  Datatype:    Char (Char)
  Dimensions:  lon × lat

julia> v[1, 2] = 'M' # OK

julia> v[1, 2, 1] = 'M' # OK

julia> v[1, 2, :] = 'M'
ERROR: BoundsError: attempt to access 4×5 NCDatasets.Variable{Char, 2, NCDataset{Nothing}} at index [1, 2, :]

julia> v[1, 2, 2] = 'M'
ERROR: BoundsError: attempt to access 4×5 NCDatasets.Variable{Char, 2, NCDataset{Nothing}} at index [1, 2, 2]
```

About the current failing test, it fails when indexing with `CartesianIndices` (or with `AbstractVector`) on `Variable` of `Char` or `String`:

```julia
julia> v
var-Char (4 × 5)
  Datatype:    Char (Char)
  Dimensions:  lon × lat
julia> ci = CartesianIndex(1,1):CartesianIndex(3,2);
julia> v.var[ci]
ERROR: MethodError: no method matching zero(::Type{Char})
Closest candidates are:
  zero(::Union{Type{P}, P}) where P<:Period at ~/julia-1.7.3/share/julia/stdlib/v1.7/Dates/src/periods.jl:53
  zero(::T) where T<:TimeType at ~/julia-1.7.3/share/julia/stdlib/v1.7/Dates/src/types.jl:450
  zero(::LinearAlgebra.UniformScaling{T}) where T at ~/julia-1.7.3/share/julia/stdlib/v1.7/LinearAlgebra/src/uniformscaling.jl:136
  ...
Stacktrace:
 [1] zeros(#unused#::Type{Char}, dims::Tuple{Int64, Int64})
   @ Base ./array.jl:526
 [2] zeros(::Type{Char}, ::Int64, ::Int64)
   @ Base ./array.jl:521
 [3] disk_getindex_batch(ar::NCDatasets.Variable{Char, 2, NCDataset{Nothing}}, indstoread::Matrix{Tuple{Int64, Int64}})
   @ DiskArrays ~/.julia/packages/DiskArrays/ivxYL/src/batchgetindex.jl:121
 [4] batchgetindex(a::NCDatasets.Variable{Char, 2, NCDataset{Nothing}}, i::CartesianIndices{2, Tuple{UnitRange{Int64}, UnitRange{Int64}}})
   @ DiskArrays ~/.julia/packages/DiskArrays/ivxYL/src/batchgetindex.jl:72
 [5] getindex_disk(a::NCDatasets.Variable{Char, 2, NCDataset{Nothing}}, i::CartesianIndices{2, Tuple{UnitRange{Int64}, UnitRange{Int64}}})
   @ DiskArrays ~/.julia/packages/DiskArrays/ivxYL/src/diskarray.jl:28
 [6] getindex(a::NCDatasets.Variable{Char, 2, NCDataset{Nothing}}, i::CartesianIndices{2, Tuple{UnitRange{Int64}, UnitRange{Int64}}})
   @ NCDatasets ~/.julia/packages/DiskArrays/ivxYL/src/diskarray.jl:178
```

I'm trying to overcome the `batchindex` fallback where we end up [here](https://github.com/meggart/DiskArrays.jl/blob/aa5343875e92b810374a20bc023a8b407002ebcc/src/diskarray.jl#L28) for eltypes `Char` and `String`.

Note that when indexing on the `CFVariable`, it works. It calls `readblock!` for each `CartesianIndex` in `ci`:
```julia
julia> v[ci]
[ Info: in readblock!. Index: (1:1, 1:1)
[ Info: in readblock!. Index: (2:2, 1:1)
[ Info: in readblock!. Index: (3:3, 1:1)
[ Info: in readblock!. Index: (1:1, 2:2)
[ Info: in readblock!. Index: (2:2, 2:2)
[ Info: in readblock!. Index: (3:3, 2:2)
3×2 Matrix{Char}:
 '\x03'  'M'
 '\x04'  '\x06'
 '\x05'  '\a'
```

My idea would be to force this behaviour for `Char` and `String`, but I don't know how.

BTW, could that be considered as a bug in DiskArrays? I suppose it would fail [here](https://github.com/meggart/DiskArrays.jl/blob/aa5343875e92b810374a20bc023a8b407002ebcc/src/batchgetindex.jl#L121) when batch indexing for any Array for which elements have no `zero` method (I haven't tried yet though). What do you think @rafaqz?

src/variable.jl
```diff
@@ -310,82 +310,61 @@ nomissing(a::AbstractArray,value) = a
...
+function readblock!(v::Variable, aout, indexes::Int...)
     datamode(v.ds)
-    return nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
+    aout .= nc_get_var1(eltype(v),v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]])
 end
 
-function Base.setindex!(v::Variable{T,N},data,indexes::Int...) where N where T
-    @debug "$(@__LINE__)"
+function readblock!(v::Variable{T,N}, aout, indexes::TR...) where {T,N} where TR <: Union{StepRange{Int,Int},UnitRange{Int}}
+    start,count,stride,jlshape = ncsub(indexes[1:N])
```
rafaqz, line 319:
```suggestion
    start,count,stride,jlshape = ncsub(indexes)
```

src/variable.jl
```diff
@@ -310,82 +310,61 @@ nomissing(a::AbstractArray,value) = a
...
-        function Base.setindex!(v::Variable{T,N},data::$data_type,indexes::StepRange{Int,Int}...) where {T,N}
-            datamode(v.ds) # make sure that the file is in data mode
-            start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
-            tmp = fill(convert(T,data),jlshape)
-            nc_put_vars(v.ds.ncid,v.varid,start,count,stride,tmp)
-            return data
-        end
-    end
+function _write_data_to_nc(v::Variable{T,N},data,indexes::Int...) where {T,N}
+    nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in indexes[ndims(v):-1:1]],T(data[1]))
```
rafaqz, line 350:
```suggestion
    nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in reverse(indexes),T(data[1]))
```

src/variable.jl
rafaqz, line 319:
```suggestion
    start,count,stride,jlshape = ncsub(indexes)
```

This is all I meant to change

src/variable.jl
rafaqz, line 350:
```suggestion
    nc_put_var1(v.ds.ncid,v.varid,[i-1 for i in reverse(indexes),T(data[1]))
```

DiskArrays.jl should take care of the `1:ndims` part? if not you would want to check those are actually trailing ones.

src/variable.jl
```diff
@@ -310,82 +310,61 @@ nomissing(a::AbstractArray,value) = a
...
-function Base.setindex!(v::Variable{T,N},data::AbstractArray{T2,N},indexes::Colon...) where {T,T2,N}
-    datamode(v.ds) # make sure that the file is in data mode
-    tmp =
-        if T <: Integer
-            round.(T,data)
-        else
-            convert(Array{T,N},data)
-        end
+function _write_data_to_nc(v::Variable{T, N}, data, indexes::StepRange{Int,Int}...) where {T, N}
+    start,count,stride,jlshape = ncsub(indexes[1:ndims(v)])
```
rafaqz, line 361:
```suggestion
    start,count,stride,jlshape = ncsub(indexes)
```

Again this should not be needed, if it is it's probably a bug in DiskArrays.jl we should fix

rafaqz:
For the `zeros` problem, in DiskArrays that liine could be something like::
```julia
    outar = OffsetArray(Array{eltype(ar),ndims(ar)}, undef, prep.outsize...),prep.offsets...)
```

See if swapping out that line in DiskArrays.jl works for you. I'm not sure where it should actually need zeros, we should be writing over those values anyway, unless I misunderstand something.

tcarion:
Thank you very much for your help @rafaqz! I integrated your suggestion in the last commit, and changing the line in DiskArrays solves the last test issue.

I think there's still an issue in DiskArrays when writing with strided indices on `Char` arrays:
```julia
julia> da = _DiskArray(fill('d', 3, 5));
julia> da[1:2:4, 1:2:5] = fill('c', 2, 3);
julia> da
3×5 _DiskArray{Char, 2, Matrix{Char}}:
 'c'                 '\x00\x00\x7f\x7e'  'c'                 '\x00\x00\x7f\x7e'  'c'
 '\x00\x00\x7f\x7e'  '\xaa\x07\x83\x60'  '\x00\x00\x7f\x7e'  '\x58\x8d\xd8\xf0'  '\x00\x00\x7f\x7e'
 'c'                 '\x00\x00\x7f\x7e'  'c'                 '\x00\x00\x7f\x7e'  'c'
```

It overwrites the non-concerned indices with undef values.

tcarion:
Yes, I get the same issue when broadcasting

rafaqz:
Ahh damn. Probably this needs a DiskArrays.jl issue.

This small array fallback code looks wrong to me, it's writing the block without first reading it to fill in the missing columns:
https://github.com/meggart/DiskArrays.jl/blob/aa5343875e92b810374a20bc023a8b407002ebcc/src/batchgetindex.jl#L196-L199

Shouldn't be hard to fix though, we just need to call `readblock!` on the temp block first.


But actually for netcdf we shouldn't use these method anyway, like I mentioned in the other comment. Because you seem to have an optimisation already in `nc_put_vars` in `_write_data_to_nc` that accepts a `stride` argument.

So instead we can just add this method (or similar):
```julia
function DiskArrays.writeblock!(v::Variable, data, indexes::AbstractVector...)
    _write_data_to_nc(v, data, indexes)
end
```

src/variable.jl
```diff
@@ -309,83 +309,59 @@ end
...
-            tmp = fill(convert(T,data),jlshape)
-            nc_put_vars(v.ds.ncid,v.varid,start,count,stride,tmp)
-            return data
-        end
-    end
+function writeblock!(v::Variable{T,N},data,indexes::AbstractUnitRange...) where {T,N}
+    datamode(v.ds)
+    _write_data_to_nc(v, data, indexes...)
+    return data
+end
```
rafaqz, line 344:
```suggestion
function writeblock!(v::Variable, data, indexes::AbstractVector...)
    _write_data_to_nc(v, data, indexes)
end
```

Sorry I missed this before - I think you want to define this for all `AbstractVector` and use your internal methods to handle them.

src/variable.jl
```diff
@@ -309,83 +309,59 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-
-function Base.getindex(v::Variable,indexes::Int...)
+function readblock!(v::Variable, aout, indexes::AbstractUnitRange...)
```
rafaqz, line 312:
```suggestion
function readblock!(v::Variable, aout, indexes::AbstractVector...)
```

tcarion:
Thanks for your advice rafaqz, it works with StepRange now!

test/test_variable_unlim.jl
```diff
@@ -39,7 +39,7 @@ defDim(ds,"lon",Inf)
 defDim(ds,"lat",110)
 v = defVar(ds,"temperature",Float32,("lon","lat"))
 data = [Float32(i+j) for i = 1:100, j = 1:110]
-v[:,1] = data[:,1]
+v[1:100,1] = data[:,1]
```
tcarion, line 42: This is also breaking the API, as @Alexander-Barth mentioned in [this comment](https://github.com/Alexander-Barth/NCDatasets.jl/pull/205#issuecomment-1512595709)

test/test_writevar.jl
```diff
@@ -26,15 +26,15 @@ for T in [UInt8,Int8,UInt16,Int16,UInt32,Int32,UInt64,Int64,Float32,Float64]
     @test all(v[:,:][:] .== 123)
 
     # write scalar
-    v[:,:] = T(100)
+    v[:,:] .= T(100)
```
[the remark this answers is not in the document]
rafaqz: Yeah that shouldn't work anyway. And without broadcast there is no chunking.

test/test_variable_unlim.jl
```diff
@@ -22,7 +22,7 @@ NCDatasets.NCDataset(filename,"c") do ds
 
         for j = 1:sz[2]
             data[:,j] .= T(j)
-            v[:,j] .= T(j)
+            v[:,j] = fill(T(j), sz[1])
```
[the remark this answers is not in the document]
rafaqz: This is where we need that `resize!` write function I've been going on about.

Unlimited dimension writing needs an API so it doesn't go through the normal `setindex!` machinery, or there will be bugs like this everywhere.

(Also from my perceptive Rasters.jl integration really needs CFVariable to be a DiskArray.jl sooner or later, or it will be lazy reads or CF standards, but not both)

test/test_variable_unlim.jl
rafaqz: I can see how that is more annoying to write now. 

Defining e.g. `v = grow!(v, :, 1, data[:, 1])` could let this work, let DiskArrays.jl fix the chunk pattern, and let Rasters.jl add a method for that and fix the lookups at the same time.

Alexander-Barth:
For the unlimlited dimensions, I would go the same route as NetCDF.jl, to write `v[1:100,1] = data[:,1]` instead of `v[:,1] = data[:,1]`. I also changed the tests in master: c56e23cc184365386375628b6f144e76835c0a49

rafaqz:
I would suggest NetCDF.jl also change that to use e.g. `grow!`, as both have broken chunking after that operation and cant be used as disk arrays.

But you know that ;)

tcarion:
I tried a very naive implementation for `grow!`, that probably won't work for all cases :-). I removed it from the test for now because it appears not to work on Julia 1.6 (I think because `replace` don't work on Tuple on this version).

Apart from this, the following failing tests appear with multi-files. I couldn't find an easy workaround so far. @rafaqz, do you know if there're existing ways of handling multi-file cases in DiskArrays?

rafaqz:
This is just an error on `show` ? 

Its coming from `size` somehow not working on the multifile netcdf:
https://github.com/Alexander-Barth/NCDatasets.jl/actions/runs/4993593286/jobs/8942938157?pr=205#step:7:151

We can probably assume that `size` should work on every variable ;)

tcarion:
Indeed, the error fails to show for some reason, making the stack trace in the tests quite misleading. I tried it outside of the test suite, and the error arises when indexing on the dimension on which the files have been concatenated:
```julia
julia> ds["var"][:,:,:,1]; # works
julia> ds["var"][:,:,:,1:2] # fails, because the 4th dimension is the one on which the multi files are concatenated
ERROR: BoundsError: attempt to access 
Stacktrace:
  [1] (::DiskArrays.var"#7#8"{NCDatasets.Variable{Float64, 3, NCDataset{Nothing}}, NTuple{4, UnitRange{Int64}}})(i::Int64)
    @ DiskArrays ~/.julia/packages/DiskArrays/639nV/src/diskarray.jl:92
  [2] foreach
    @ ./abstractarray.jl:2712 [inlined]
  [3] interpret_indices_disk(A::NCDatasets.Variable{Float64, 3, NCDataset{Nothing}}, r::NTuple{4, UnitRange{Int64}})
    @ DiskArrays ~/.julia/packages/DiskArrays/639nV/src/diskarray.jl:91
  [4] getindex_disk(::NCDatasets.Variable{Float64, 3, NCDataset{Nothing}}, ::UnitRange{Int64}, ::Vararg{UnitRange{Int64}})
    @ DiskArrays ~/.julia/packages/DiskArrays/639nV/src/diskarray.jl:29
  [5] getindex
    @ ~/.julia/packages/DiskArrays/639nV/src/diskarray.jl:178 [inlined]
  [6] (::NCDatasets.var"#86#87"{NTuple{4, UnitRange{Int64}}})(v::NCDatasets.Variable{Float64, 3, NCDataset{Nothing}})
    @ NCDatasets ~/.julia/dev/NCDatasets/src/defer.jl:126
  [7] (::NCDatasets.var"#84#85"{NCDatasets.var"#86#87"{NTuple{4, UnitRange{Int64}}}, NCDatasets.DeferVariable{Float64, 3}})(ds::NCDataset{Nothing})
    @ NCDatasets ~/.julia/dev/NCDatasets/src/defer.jl:104
  [8] NCDataset(::NCDatasets.var"#84#85"{NCDatasets.var"#86#87"{NTuple{4, UnitRange{Int64}}}, NCDatasets.DeferVariable{Float64, 3}}, ::String, ::Vararg{String}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ NCDatasets ~/.julia/dev/NCDatasets/src/dataset.jl:255
  [9] NCDataset
    @ ~/.julia/dev/NCDatasets/src/dataset.jl:253 [inlined]
 [10] Variable
    @ ~/.julia/dev/NCDatasets/src/defer.jl:103 [inlined]
 [11] getindex
    @ ~/.julia/dev/NCDatasets/src/defer.jl:125 [inlined]
 [12] getindex(::CommonDataModel.CFVariable{Union{Missing, Float64}, 3, NCDatasets.DeferVariable{Float64, 3}, NCDatasets.DeferAttributes, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float64, Tuple{}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::UnitRange{Int64}, ::UnitRange{Int64})
    @ CommonDataModel ~/.julia/dev/CommonDataModel/src/cfvariable.jl:389
 [13] getindex(::NCDatasets.CatArrays.CatArray{Union{Missing, Float64}, 4, 3, CommonDataModel.CFVariable{Union{Missing, Float64}, 3, NCDatasets.DeferVariable{Float64, 3}, NCDatasets.DeferAttributes, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float64, Tuple{}, Nothing, Nothing, Nothing, Nothing, Nothing}}}}, ::Function, ::Function, ::Function, ::UnitRange{Int64})
    @ NCDatasets.CatArrays ~/.julia/dev/NCDatasets/src/CatArrays.jl:79
 [14] getindex(::NCDatasets.MFCFVariable{Union{Missing, Float64}, 4, 3, CommonDataModel.CFVariable{Union{Missing, Float64}, 3, NCDatasets.DeferVariable{Float64, 3}, NCDatasets.DeferAttributes, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Float64, Tuple{}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, NCDatasets.MFVariable{Float64, 4, 3, NCDatasets.DeferVariable{Float64, 3}, NCDatasets.DeferAttributes, NCDatasets.MFDataset{DeferDataset, 1, String, NCDatasets.DeferAttributes, NCDatasets.DeferDimensions, NCDatasets.DeferGroups}}, NCDatasets.DeferAttributes, NCDatasets.MFDataset{DeferDataset, 1, String, NCDatasets.DeferAttributes, NCDatasets.DeferDimensions, NCDatasets.DeferGroups}}, ::Function, ::Function, ::Function, ::UnitRange{Int64})
    @ NCDatasets ~/.julia/dev/NCDatasets/src/multifile.jl:293
 [15] top-level scope
    @ REPL[110]:1
SYSTEM (REPL): showing an error caused an error
```

You can see here that the error happens somewhere in DiskArrays

tcarion:
I suspect it to be a DiskArrays bug.
In standard Julia arrays, you can index an `Array{T,N}` on the N+1 dimension with `1:1`. This returns an array with size `(size(A)..., 1)` :
```julia
julia> A = rand(2, 3);
julia> A[:,:,1:1]
2×3×1 Array{Float64, 3}:
[:, :, 1] =
 0.745692  0.288329  0.071235
 0.613755  0.240925  0.826111
```

Doing the same with a DiskArray fails:
```julia
julia> DA =_DiskArray(A);
julia> DA[:,:,1:1]
ERROR: BoundsError: attempt to access 2×3 _DiskArray{Float64, 2, Matrix{Float64}} at index [:, :, 1:1]
Stacktrace:
 [1] (::DiskArrays.var"#7#8"{_DiskArray{Float64, 2, Matrix{Float64}}, Tuple{Colon, Colon, UnitRange{Int64}}})(i::Int64)
   @ DiskArrays ~/.julia/dev/DiskArrays/src/diskarray.jl:92
 [2] foreach
   @ ./abstractarray.jl:2712 [inlined]
 [3] interpret_indices_disk(A::_DiskArray{Float64, 2, Matrix{Float64}}, r::Tuple{Colon, Colon, UnitRange{Int64}})
   @ DiskArrays ~/.julia/dev/DiskArrays/src/diskarray.jl:91
 [4] getindex_disk(::_DiskArray{Float64, 2, Matrix{Float64}}, ::Function, ::Vararg{Any})
   @ DiskArrays ~/.julia/dev/DiskArrays/src/diskarray.jl:29
 [5] getindex(::_DiskArray{Float64, 2, Matrix{Float64}}, ::Function, ::Function, ::UnitRange{Int64})
   @ Main ~/.julia/dev/DiskArrays/src/diskarray.jl:178
 [6] top-level scope
   @ REPL[26]:1
```

@rafaqz, do you agree that the DiskArray should behave like a regular array here? Thanks!

rafaqz:
Yep that should work. Basically all base array behaviours should work.

Make an issue ;)

tcarion:
The tests are now passing with the new version of DiskArrays

src/NCDatasets.jl
```diff
@@ -65,6 +68,8 @@ include("ncgen.jl")
 include("select.jl")
 include("precompile.jl")
 
+@implement_diskarray NCDatasets.Variable
```
rafaqz, line 71: Are we aiming to move more of these to DiskArrays.jl?

Just having `Variable` means this PR doesn't allow Rasters.jl integration of gribb or netcdf.

src/variable.jl
```diff
@@ -309,85 +309,73 @@ end
...
+    inds = replace(indexes, Colon() => 1:length(data))
+    v[inds...] = data
+    return v
 end
 
+getchunksize(v::Variable) = getchunksize(haschunks(v),v)
+getchunksize(::DiskArrays.Chunked, v::Variable) = chunking(v)[2]
+# getchunksize(::DiskArrays.Unchunked, v::Variable) = DiskArrays.estimate_chunksize(v)
+getchunksize(::DiskArrays.Unchunked, v::Variable) = size(v)
+eachchunk(v::CFVariable) = eachchunk(v.var)
```
rafaqz, line 374: These are defined on `CFVariable`, but it's not implementing the rest of the DiskArrays.jl interface

src/variable.jl
```diff
@@ -309,85 +309,73 @@ end
...
+end
 
-    nc_put_var(v.ds.ncid,v.varid,tmp)
-    return data
+function grow!(v::CFVariable, data, indexes::Union{Integer, Colon}...)
+    unlimdims = unlimited(Dimensions(CommonDataModel.dataset(v)))
+    length(unlimdims) == 1 || error("Only 1 unlimited dimension is supported") 
+    alldims = dimnames(v)
+    iunlimdim = findfirst(==(unlimdims[1]), alldims)
+    icol = findfirst(x -> x isa Colon, indexes)
```
rafaqz, line 363: What if there are other colons? shouldn't you just check that the `iunlimdim` index is a `Colon`

src/NCDatasets.jl
tcarion: I'm experimenting with this on this PR: JuliaGeo/CommonDataModel.jl/pull/9.

The issue I'm currently facing is, I think, related to the growth of unlimited dimensions. As you predicted, this causes problems when CFVariable implements DiskArray.

Unfortunately, I can't show you the error stack on github since it needs the combination of this branch with JuliaGeo/CommonDataModel.jl/pull/9. Is there a way to tell github to run CI with a specific branch of a dependency?

src/variable.jl
tcarion: You're right, I must admit I implemented this pretty rapidly, and I didn't really test it. But I guess we'll have to look more deeply into it given the issue above!

src/NCDatasets.jl
rafaqz: You can add a Project.toml and Mainfest.toml to /test if you need too, but its a bit of fiddling and maybe no better than just pasting errors in issues.

Yes that is fairly predictable. You will need to rebuild the chunking after `grow!`.

src/variable.jl
rafaqz: As you saw with the bugs in DiskArrays edge cases its actually pretty hard to do it properly.

Alexander-Barth:
I just updated the small benchmark numbers:
https://github.com/Alexander-Barth/NCDatasets.jl/commit/8662c883a9720c0831d19a0545b68739b8651ff7

using the last released version of NCDatasets (we got faster 🎉, R-ncdf4 too).
I am wondering what the impact of this PR is...

tcarion:
I run the benchmark:

**NCDatasets#master:**
| Module           | median | minimum |  mean | std. dev. |
|:---------------- | ------:| -------:| -----:| ---------:|
| R-ncdf4          |  0.374 |   0.350 | 0.375 |     0.012 |
| python-netCDF4   |  0.622 |   0.609 | 0.634 |     0.028 |
| julia-NCDatasets |  0.355 |   0.312 | 0.361 |     0.028 |

**this branch:**
| Module           | median | minimum |  mean | std. dev. |
|:---------------- | ------:| -------:| -----:| ---------:|
| R-ncdf4          |  0.416 |   0.359 | 0.424 |     0.044 |
| python-netCDF4   |  0.623 |   0.609 | 0.642 |     0.040 |
| julia-NCDatasets |  0.236 |   0.190 | 0.238 |     0.019 |

Unfortunately, it appears the performances are impacted...

rafaqz:
What is it benchmarking though?

Like how big is the file, how is it chunked? And what is actually different? 

I would guess it may be now doing multiple reads of a chunked file where earlier it was doing just one read?

Finally, how do we use this information to improve DiskArrays.jl generally.

There will always be some overhead to allowing larger than memory reads and writes, but we can minimise it if we understand the problem.

meggart:
Yes, could you point to the benchmark code so we can see what exactly is benchmarked? I would be very interested to see where the regression comes from.

Alexander-Barth:
Thanks a lot @tcarion for running the benchmarks! (I hope you did not have to install R and python just for this 😅 )

The benchmark code is this one:
https://github.com/Alexander-Barth/NCDatasets.jl/blob/8662c883a9720c0831d19a0545b68739b8651ff7/test/perf/benchmark-julia-NCDatasets.jl

Essentially:
> The benchmark loads a variable of the size 1000x500x100 in slices of 1000x500 (applying the scaling of the CF conventions) and computes the maximum of each slice and the average of each maximum over all slices. This operation is repeated 100 times. 

The file is generated from this scipt:
https://github.com/Alexander-Barth/NCDatasets.jl/blob/master/test/perf/generate_data.jl

Concerning the cache in DiskArrays, at which point is the memory released? Once the variable goes out-of-scope or once the netcdf file is closed? Is the cache per variable or a global cache?
If the data on this file is changed, is there a risk that cached / out-dated data is returned?

rafaqz:
There is no caching of data in DiskArrays

meggart:
As a first test I ran the performance tests for NetCDF.jl, which is based on DiskArrays as well and the results are very comparable to NCDatasets, so I think in theory it should be possible to make a fast implementation using DiskArrays:

I ran this code with NCDatasets#master and the tagged version of NetCDF.jl:

````julia

using NCDatasets
using Statistics
using BenchmarkTools

function compute(v)
    tot = zero(eltype(v))
    for n = 1:size(v,3)
        slice = v[:,:,n]
        partial_sum = maximum(skipmissing(slice))
        tot += partial_sum
    end
    return tot/size(v,3)
end

function process(fname)
    ds = NCDataset(fname,"r") do ds
        v = ds["v1"];
        tot = compute(v)
        return tot
    end
end


tot = process(fname)

using NetCDF, DiskArrayTools
function processnc(fname)
    ds = NetCDF.open(fname) do ds
        v = CFDiskArray(ds["v1"],ds["v1"].atts);
        tot = compute(v)
        return tot
    end
end

fname = "filename_fv.nc"

tot = process(fname)
tot = processnc(fname)

bm = run(@benchmarkable process(fname) samples=100 seconds=10000)

@show bm

bmnc = run(@benchmarkable processnc(fname) samples=100 seconds=10000)

@show bmnc
````

and got for NCDatasets#master:

````
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  349.407 ms … 603.973 ms  ┊ GC (min … max): 11.11% … 9.96%
 Time  (median):     389.497 ms               ┊ GC (median):    11.20%
 Time  (mean ± σ):   407.478 ms ±  52.350 ms  ┊ GC (mean ± σ):  11.29% ± 1.08%

  ▂ ▄▆▆ ▄▄██▄▄     ▄    
  ████████████▆█▆▁▁██▆▁▆▆▆▄▁▆█▁▁▄▄▄▁▁▄▄▁▁▁▄▁▁▁▁▁▄▁▁▁▁▁▁▁▁▄▁▁▁▁▄ ▄
  349 ms           Histogram: frequency by time          604 ms <

 Memory estimate: 476.91 MiB, allocs estimate: 1824.
````

and for NetCDF.jl:

````
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  336.217 ms … 641.709 ms  ┊ GC (min … max): 12.18% … 9.80%
 Time  (median):     370.787 ms               ┊ GC (median):    11.50%
 Time  (mean ± σ):   382.203 ms ±  48.168 ms  ┊ GC (mean ± σ):  11.57% ± 0.65%

  ▅ █▆▂  ▂▆▃▃▂ ▂
  █▇██████████▇█▅▅▄▄▁▅▄█▄▁▁▁▁▁▄▄▁▁▄▄▁▄▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▄ ▄
  336 ms           Histogram: frequency by time          587 ms <

 Memory estimate: 477.01 MiB, allocs estimate: 5540.
````

I will test against this branch next...

**Comment by meggart** [comment]
Ok, I see the regression as well, median goes down to 560ms for this branch on my (slow) machine. I will try to look at the code and see what the main difference compared to the NetCDF.jl DiskArray implementation could be.

File: src/variable.jl (hunk 2 above)
Comment (meggart) [reply src/variable.jl:350]: I think by defining `readblock!` and `writeblock!` for StepRanges should be sufficient to go into these optimisations. DiskArrays does provide some fallback routines in case only `AbstractUnitRange`s are supported by the backend but they should not be called in this case.

**Comment by tcarion** [comment]
Thanks a lot for taking a look! My implementation mainly was trial and error, and I didn't always completely understand what was behind... So I might have done falsy and naive stuff :-)

**Comment by Alexander-Barth** [comment]
> There is no caching of data in DiskArrays

Are you sure? What about this [README.md](https://github.com/meggart/DiskArrays.jl/blob/main/README.md):
> an iterator over the values of a DiskArray that caches a chunk of data and returns the values within. This allows efficient usage of e.g. using DataStructures; counter(a)

**Comment by rafaqz** [comment]
It doesnt keep it afterwards though, only dyring iteration an droadcast over the chunks. Its very short term caching.

**Comment by meggart** [comment]
> Thanks a lot for taking a look! My implementation mainly was trial and error, and I didn't always completely understand what was behind... So I might have done falsy and naive stuff :-)

Ok after some digging I think I found it. There seems to be a dispatch problem, so that a slow fallback method is called instead of your `readblock!`. Just try

````julia
fname="filename_fv.nc"

ds = NCDataset(fname,"r")
v = ds["v1"];
aout = zeros(UInt8,100,100)
@which DiskArrays.readblock!(v.var,aout,Base.OneTo(100),Base.OneTo(100),1:1)
````
this calls the following fallback:
````
readblock!(A::NCDatasets.Variable, A_ret, r::AbstractVector...)
     @ NCDatasets C:\Users\fgans\.julia\dev\DiskArrays\src\batchgetindex.jl:208
````

although IMO a more specific method is defined:

````julia
methods(DiskArrays.readblock!)
````
````
 [7] readblock!(v::NCDatasets.Variable, aout, indexes::TI...) where TI<:Union{AbstractUnitRange, StepRange}
     @ NCDatasets C:\Users\fgans\.julia\dev\NCDatasets\src\variable.jl:313
 [8] readblock!(A::NCDatasets.Variable, A_ret, r::AbstractVector...)
     @ NCDatasets C:\Users\fgans\.julia\dev\DiskArrays\src\batchgetindex.jl:208
````

I have no idea why it is calling [8] instead of [7]

**Comment by tcarion** [comment]
That's weird because when running it, it uses [7]:
```julia
function readblock!(v::Variable, aout, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
    datamode(v.ds)
    @info "in specialized readblock, indices=$indexes"
    _read_data_from_nc!(v, aout, indexes...)
    return aout
end

julia> DiskArrays.readblock!(v.var,aout,Base.OneTo(100),Base.OneTo(100),1:1)
[ Info: in specialized readblock, indices=(1:100, 1:100, 1:1)
```

Wouldn't that rather be a bug in the `@which` macro?

meggart:
At one point it is going into that function, but it is taking a de-tour through a DiskArrays fallback. Just replace your `@info` macro with throwing a error and you will see a stacktrace.

tcarion:
You're right, sorry!
I found a workaround by defining two methods for each index type instead of using an Union (`readblock!(v::Variable, aout, indexes::AbstractUnitRange...)` and `readblock!(v::Variable, aout, indexes::StepRange...)` instead of `readblock!(v::Variable, aout, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}`)

Doing that, I get:
```julia
julia> @which DiskArrays.readblock!(v.var,aout,Base.OneTo(100),Base.OneTo(100),1:1)
readblock!(v::NCDatasets.Variable, aout, indexes::AbstractUnitRange...)
     @ NCDatasets ~/.julia/dev/NCDatasets/src/variable.jl:312
```

I don't know if that is how Julia is supposed to work, but, IMO, it's pretty unintuitive. And this implies duplicating the code for `readblock!`. I guess the same applies to `writeblock!`. 

I'll try to run the performance tests with this fix

tcarion:
It seems that was the problem! Thanks @meggart !

| Module           | median | minimum |  mean | std. dev. |
|:---------------- | ------:| -------:| -----:| ---------:|
| before fix |  0.387 |   0.350 | 0.393 |     0.023 |
| after fix |  0.215 |   0.179 | 0.218 |     0.015 |

rafaqz:
You don't have to duplicate code, you can just put it in another function `_readblock!` and call it from both `readblock!` methods. Its not unusual to have to do a few things like this to get dispatch working how you want.

tcarion:
Alright, thanks for the tip, I'll do that

Alexander-Barth:
Thank you all for you help and explanations here!
Let me know once you feel that it should be merged :-)
It should probably be version 0.13 ...

tcarion:
We need to decide whether we want the DiskArray implementation for `CDM.CFVariable` with this PR or not. If so, we need to settle on the API for growing a variable on some dimension, as discussed above. So basically, I see two options:
1. We merge this PR, and we'll make another PR for `CDM.CFVariable` as DiskArray
2. We work on `CDM.CFVariable` as DiskArray here, and merge once everything is done.

I would personally go for 2., so it induces breaking changes to NCDatasets only once :-)

rafaqz:
It would be nice to add `grow!` to DiskArrays and extend it everywhere else. Then other packages like Raaters can also extend `grow!`. 

This actually came up today: we mostly have to write new tiff slices to separate files when builing larger-than-memory datasets with Rasters.jl, because theres no interface for growing netcdf. But writing time slices to netcdf makes more sense when its really one dataset.

It would also be good to have a `setgrowdimension!` or something in CommonDataModel.jl so we can do that easily.

rafaqz:
It would be nice to add `grow!` to DiskArrays and extend it everywhere else. Then other packages like Raaters can also extend `grow!`. 

This actually came up today: we mostly have to write new tiff slices to separate files when builing larger-than-memory datasets with Rasters.jl, because theres no interface for growing netcdf. But writing time slices to netcdf makes more sense when its really one dataset.

It would also be good to have a `setgrowdimension!` or something in CommonDataModel.jl or DiskArrays.jl so we can do that easily.

Alexander-Barth:
As I mentioned before, I don't like the `grow!` function because it is not easy to use and not very intuitive in my opinion. As far as I am aware, there is no such equivalent function in python's netCDF4, matlab netcdf and octave netcdf toolboxes. Also the C-API (and C++, Fortran too) does not make any difference between between write to a fixed sized array and writing to an array with growable size. (In netCDF4 there can also be multiple unlimited dimensions, but you know that :-))

@tcarion is it possible for you to remove that `grow!`  function from the PR?

rafaqz:
@Alexander-Barth the problem is via CommonDataModel.jl you have now coupled your specific wants to multiple other packages that need `CFVariable` to be a `DiskArray`.

You know there is no way of using `setindex!` how you want without breaking DiskArrays and Julia base array behaviour.

Other languages do not have this problem because they dont have the luxury of treating any objects as generic arrays that arbitrary methods from users and unrelated packages will work on. 

We do have that luxury, and thus we also have this problem. 

What do you suggest we do instead. This set of PRs has so far cost a lot of our dev hours and attention, and we really need to get something usable at the end.

meggart:
I don't think I can contribute a lot to solve this discussion, just as a demonstration, for Zarr.jl I took the freedom to give `append!` a `dims` keyword argument https://github.com/JuliaIO/Zarr.jl/blob/4faceed95d5ba2a800249e3881784ed7f4da4014/src/ZArray.jl#L424-L458C4 . This way we did not have to export a new function name and growing arrays has it interface.

rafaqz:
`append!` would also work

rafaqz:
@tcarion another option is we dont worry about CFVariable and implement DiskArrays compat for CFVariable and `grow!` or whatever in Rasters.jl.

We would need to add a wrapper `Rasters.CFVariable` so its not type piracy, but then it can work.

This is not amazing design but I have put a lot of time into this over the last year with nothing useable in Rasters.jl so far, and it would be a quick win to get gribb support and proper DiskArrays support for NCDatasets.

tcarion:
I'm definitely less experienced with Julia compared to you guys, so I might not see the big picture, but I understand it's a complicated choice with good points and drawbacks on both sides. I think I would prefer to stick with the standard Julia array behaviour, but it's also cumbersome to break a quite convenient API.

I'll remove the ´grow!´ function here for now and I'll try @rafaqz's approach [here](https://github.com/rafaqz/Rasters.jl/pull/416)

tcarion:
I had to add some tweaks when writing because of the new way of growing a variable ([see here]). 

@ Alexander-Barth, is it ok for you to merge?

rafaqz:
@Alexander-Barth can we get this merged?

Alexander-Barth:
Sorry I did not see your message (I got a lot email from github recently that I missed) that you are ready with the PR.
I think this PR depends on https://github.com/JuliaGeo/CommonDataModel.jl/pull/9. Is this ready too?
For NCDatasets, I will draft some release notes to document the changes. Maybe you will have a time to review it.

Alexander-Barth:
For the release notes:
The array API of NCDatasets is now more similar to base Julia in particular:

* `ncvar[indices] = scalar` should now be  `ncvar[indices] .= scalar`
* `ncvar2D[:]` flattens the data in the 2D NetCDF variable `ncvar2D`. To read the full array one need to use `ncvar2D[:,:]` or `Array(ncvar2D)` (similarly for 3D, 4D... arrays).

rafaqz:
This doesn't depend on CommonDataModel changes, as we cant do this for `AbstractVariable` currently (unless you have changed your mind). 

These are kind of demonstration PRs... ultimately we need to make `AbstractVariable <: AbstractDiskArray` in CommonDataModel.jl for much of this to be usable, and avoid duplicating all of this code everywhere. 

It would fix issues like this and allow deleting `SubVariable` and `CatArray`, and probably a lot more code here and in CommonDataModel. https://github.com/Alexander-Barth/NCDatasets.jl/issues/219

But it will hit your `setindex!` problem somewhere or other.

I'm happy to do this work if its something you want. Otherwise merging this is useful as a stopgap so we can hack together temporary DiskArrays support elsewhere on top of `Variable`. This is all pretty horrible though, we relaly just need `AbstractVariable <: AbstractDiskArray`

tcarion:
I confirm that https://github.com/JuliaGeo/CommonDataModel.jl/pull/9 is not needed for this PR to work. https://github.com/JuliaGeo/CommonDataModel.jl/pull/9 was intended to try to make all `AbstractVariable <: AbstractDiskArray`, which I think would be the best long-term solution, but, as @rafaqz mentioned, is failing because of the unlimited dimension issue. 

The release notes seem nice and I think they cover everything that has changed, thank you for writing them!

Alexander-Barth:
Unfortunately, I see some slow down for the benchmark on my machine. Current master runs in 181.193 ms  while this PR 307.836 ms. Do you see this as well? I am using DiskArrays v0.3.14 (current version).

I tried this PR with another package of our package (DIVAnd) and upgrade was quite smooth.

For info: this commit  7edf68f7891fb28dba5a92acc6cbd702509f386e disable the opendap tests which causes current test failures (opendap server seems to be down).


Master
```
julia> bm
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  162.789 ms … 231.338 ms  ┊ GC (min … max): 1.82% … 15.08%
 Time  (median):     181.193 ms               ┊ GC (median):    2.68%
 Time  (mean ± σ):   183.786 ms ±  16.540 ms  ┊ GC (mean ± σ):  3.09% ±  1.54%

  █ ▄▂▂            ▂▆  ▂ ▂ ▂                                     
  █▆███▆▆▁▄▆▆▆▆█▄▄███▄▄█▄█▁███▆▆▆▆▁▆▄▁▄▁▁▁▄▄▁▁▄▄▁▄▁▁▁▁▁▁▆▄▄▄▆▁▄ ▄
  163 ms           Histogram: frequency by time          224 ms <

 Memory estimate: 476.90 MiB, allocs estimate: 1610.

```

PR
```
julia> bm
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  291.086 ms … 326.860 ms  ┊ GC (min … max): 1.15% … 1.50%
 Time  (median):     307.836 ms               ┊ GC (median):    1.74%
 Time  (mean ± σ):   307.185 ms ±   4.399 ms  ┊ GC (mean ± σ):  1.74% ± 0.22%

                            ▃   ▃ ▂     ▅▂▅ █▃                   
  ▄▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▄▁▁▄▁▅▄▅▄█▁█▇█▄█▅▄▄▇▇███▇████▅▄▇▄▄▁▁▁▁▁▁▁▁▁▄ ▄
  291 ms           Histogram: frequency by time          318 ms <

 Memory estimate: 527.15 MiB, allocs estimate: 8010.
```

src/variable.jl
```diff
@@ -345,85 +345,68 @@ end
 nomissing(a::AbstractArray,value) = a
 export nomissing
 
-
-function Base.getindex(v::Variable,indexes::Int...)
+# This method needs to be duplicated instead of using an Union. Otherwise a DiskArrays fallback is called instead which impacts performances 
+# (see https://github.com/Alexander-Barth/NCDatasets.jl/pull/205#issuecomment-1589575041)
+function readblock!(v::Variable, aout, indexes::TI...) where TI <: Union{AbstractUnitRange,StepRange}
```
rafaqz, line 350:
```suggestion
function readblock!(v::Variable, aout, indexes::Union{<:AbstractUnitRange,<:StepRange}...)
```

This dispatch was not correct so we were using the fallback for `AbstractVector`.

I get roughly the same performance as `master` after this change.

rafaqz:
@Alexander-Barth see the fix above for your performance question.

Took a few hours to track down :sweat_smile: - that's a very subtle bug in dispatch on `readblock!` where using a type parameter for mixed type varargs its not the same as using the types directly in the signature.

Hopefully this is good to merge now.

rafaqz:
@Alexander-Barth this shouldnt have a warning now

Alexander-Barth:
With DiskArrays, v0.3.18, there seam to be macro issue:

Line 71 on NCDatasets.jl is `@implement_diskarray NCDatasets.Variable`.

Any idea which is could be the case?

```
julia> using NCDatasets
[ Info: Precompiling NCDatasets [85f8d34a-cbdd-5861-8df4-14fed0d494ab]
ERROR: LoadError: syntax: "NCDatasets.Variable" is not a valid function argument name around /home/abarth/.julia/packages/DiskArrays/Wl1VA/src/zip.jl:77
Stacktrace:
 [1] top-level scope
   @ ~/.julia/dev/NCDatasets/src/NCDatasets.jl:71
 [2] include
   @ ./Base.jl:457 [inlined]
 [3] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2010
 [4] top-level scope
   @ stdin:2
in expression starting at /home/abarth/.julia/dev/NCDatasets/src/NCDatasets.jl:1
in expression starting at stdin:2
```

rafaqz:
Feel free to PR your fix ;)

rafaqz:
Ok fixed, will be registered soon. That method is only for preventing ambiguity before throwing an error, so isn't caught in the DiskArrays tests because we don't use the macro on `Package.ArrayType` but just `ArrayType`.

Anyway, there are some overheads to using these macros rather than inheritance, we were actually close to deleting the macros in DiskArrays.jl before this PR.

rafaqz:
@Alexander-Barth this should be good to merge

Alexander-Barth:
I still see a method warning:
```
WARNING: Method definition _reshape(DiskArrays.AbstractDiskArray{var"#s69", 1} where var"#s69", Tuple{Int64}) in module DiskArrays at /home/abarth/.julia/packages/DiskArrays/fshK3/src/reshape.jl:83 overwritten in module NCDatasets on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **
```

But with these two commits https://github.com/Alexander-Barth/NCDatasets.jl/commit/08a4888c4e73220f6df2a33fb96a7a50ee0515f0 and https://github.com/Alexander-Barth/NCDatasets.jl/commit/7edf68f7891fb28dba5a92acc6cbd702509f386e (unrelated to DiskArrays) all NCDatasets test pass.

Can this warning be fixed before I merge? Thanks a lot!

tcarion:
Thank you both for your help! Sorry I'm not very available lastly...
I also get this warning coming from DiskArrays.jl

rafaqz:
Weird, Im not getting that locally or in CI. Will try to reproduce

rafaqz:
Ok, this was there for an ambiguity method. Will be fixed soon.

Unfortunately we cant fix the ambiguity for NCDatasets.Variable in the macro because we cant know the type parameters of arbitrary types. Using the macro makes these things difficult to handle, really we shouldn't use them.

rafaqz:
The last fix is registered, hopefully that does it

Alexander-Barth:
Thanks a lot! Everything works now fine too on my side.

rafaqz:
Amazing so excited to have the basics of this working now.

rafaqz:
Can we get a new version tagged so we can work on https://github.com/rafaqz/Rasters.jl/pull/416?

Alexander-Barth:
In my tests, unfortunately the slow `batchgetindex` is still called when using `v[:,:,1]` with DiskArray 0.3.22.

I traced it down how julia handles function specificity. Surprisingly we get:

```
julia> foo(i::AbstractVector...) = "vectors"
foo (generic function with 1 method)

julia> foo(i::TI...) where TI <: Union{AbstractUnitRange,StepRange} = "range"
foo (generic function with 2 methods)

julia> foo(Base.OneTo(1),Base.OneTo(1),1:1)
"vectors"
```

The fix is to define the following function:
```
julia> foo(i::AbstractRange...) = "range 2"
foo (generic function with 3 methods)

julia> foo(Base.OneTo(1),Base.OneTo(1),1:1)
"range 2"
```

Unfortunately, even avoiding batchgetindex `v[:,:,1]` is still not as fast as in NCDataset 0.12 without DiskArray:

NCDatasets 0.13.1 with DiskArrays

```
julia> v1 = @btime v[:,:,1];
  35.120 μs (36 allocations: 254.42 KiB)

julia> @which v[:,:,1]
getindex(a::NCDatasets.Variable, i...)
     @ NCDatasets ~/.julia/dev/DiskArrays/src/diskarray.jl:215
```

NCDatasets 0.12.17 without DiskArrays

```

julia> v1 = @btime v[:,:,1];
  23.730 μs (8 allocations: 253.45 KiB)

julia> @which v[:,:,1]
getindex(v::NCDatasets.Variable{T, N}, indexes::Union{Colon, Int64, AbstractRange{<:Integer}}...) where {T, N}                               
     @ NCDatasets ~/.julia/packages/NCDatasets/st9Jz/src/variable.jl:482

```

Any ideas how we can improve would be appreciated :-)
@tcarion @meggart @rafaqz

PR #221, opened 2023-07-29
Swap AbstractVariable to SubVariable

### LINKED ISSUE #219: Type piracy on CommonDataModel
This line is type piracy on CommonDataModel.jl:

https://github.com/Alexander-Barth/NCDatasets.jl/blob/66d1e672ad5419ac5af96ce022c5904e678fdd30/src/subvariable.jl#L94

Which causes test failures here: https://github.com/rafaqz/Rasters.jl/actions/runs/5475178088/job/14825206026?pr=416#step:5:802

@tcarion FYI

Should it instead dispatch on `SubVariable` ?

rafaqz:
Fixes #219

### DIFF
--- a/src/subvariable.jl
+++ b/src/subvariable.jl
@@ -66,7 +66,7 @@ function Base.view(v::SubVariable,indices::Union{Int,Colon,AbstractVector{Int}}.
 end
 
 """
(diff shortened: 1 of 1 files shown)


rafaqz:
Seems that this type piracy is actually required. Should these `view` methods be in CommonDataModel.jl?

PR #236, opened 2023-11-01
Minor docs improvements

glwagner:
Just a little extra formatting plus cross-reference to docs for `cfvariable`. (Found while looking for docs on `DateTimeNoLeap`.)


Alexander-Barth:
Thanks a lot😀

PR #239, opened 2023-11-26
Corrected the SST image output

keduba:
I changed the CMEMS output file  `example_SST_CMEMS.png` to `example_SST_CMEMS_fix.png`. The first one had the string "time" in the title instead of the actual time of the file "2008-01-01T00:00:00".

Then I modified the tutorial image alt-text and link to point to the new file.


Alexander-Barth:
Thanks, Kene, but the file is actually automatically generated by running the code. Maybe there is conflict with the julia function `time` and the `time` variable  need to be renamed (e.g. `SST_time`) ?

keduba:
Okay. So I'll just test it with reverting and renaming the `time` variable and see if it works.

keduba:
> Okay. So I'll just test it with reverting and renaming the `time` variable and see if it works. I'm also assuming the actual file that is run is a `.jl` file not the `tutorials.md`.

Never mind. I realised it's quoted in a code block. I've effected the change and will submit a new PR. Thank you.

Alexander-Barth:
Thanks!

Alexander-Barth:
Thank for the fix!

keduba:
Hi Alex, I just checked. The generated plot still has the same issue. Could it be that it's just pointing to the previous plot and not generating it afresh ?

Alexander-Barth:
Thanks a lot for letting me know. I just fixed it manually.
https://github.com/Alexander-Barth/NCDatasets.jl/commit/9f198d064cc22701ee94a56f58bac8a88b992302

PR #243, opened 2023-12-19
Remove references to obsolete types

Sbozzolo:
`NCDatasets.Dimensions` is no longer around as far as I can tell.

### DIFF
--- a/docs/src/dimensions.md
+++ b/docs/src/dimensions.md
@@ -4,11 +4,9 @@ In the NetCDF data model, dimensions have names and a length (but possibly an un
 For a given `Variable` or `CFVariable`,the names of the corresponding dimensions are obtained with using [`dimnames`](@ref).
 
 ```@docs
-keys(d::NCDatasets.Dimensions)
 haskey(a::NCDatasets.NCIterable,name::AbstractString)
 defDim
-unlimited(d::NCDatasets.Dimensions)
-setindex!(d::NCDatasets.Dimensions,len,name::AbstractString)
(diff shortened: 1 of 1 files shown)


Alexander-Barth:
It moved to `CommonDataModel`, but it is imported in `NCDatasets` from `CommonDataModel`.
The type `NCDatasets.Dimensions` is still defined (but as `CommonDataModel.Dimensions`)
 
The functions that you want to remove still work as far as I can tell.

```
julia> using NCDatasets

julia> NCDatasets.Dimensions
CommonDataModel.Dimensions

julia> ds = NCDataset("/tmp/foo.nc","c"); defVar(ds,"data",randn(4,3),("lon","lat"),attrib=Dict("foo" => "bar")); close(ds);

julia> ds = NCDataset("/tmp/foo.nc","a");

julia> unlimited(ds.dim)
String[]

julia> keys(ds.dim)
2-element Vector{String}:
 "lon"
 "lat"

julia> setindex!(ds.dim,123,"time")
```

Can you provide a MWE if they do not work for you?

Sbozzolo:
I see this in the documentation:





https://alexander-barth.github.io/NCDatasets.jl/stable/dimensions/

Alexander-Barth:
OK, I see now! I forgot to add the docstring to the new functions in CommonDataModel which replaced the old ones. I added them to CommonDataModel and they are now also available as `keys(d::NCDatasets.Dimensions)` (and similar).

The documentation should now be fixed in the dev version.

The PR should not be necessary, but thank you for your time.

PR #255, opened 2024-04-09
Throw informative error for unsupported vtype

Sbozzolo:
Currently, defVar fails with something like
```
ERROR: KeyError: key DateTime not found
```
This PR throws a more informative error message


Alexander-Barth:
Thanks a lot @Sbozzolo !

PR #276, opened 2025-03-24
Declare compatibility with netCDF 4.9.3

visr:
Like https://github.com/JuliaGeo/NetCDF.jl/pull/201, but using the exact match compat like is done here.

https://github.com/JuliaBinaryWrappers/NetCDF_jll.jl/releases/tag/NetCDF-v401.900.300%2B0
https://github.com/JuliaPackaging/Yggdrasil/pull/10636
https://github.com/JuliaPackaging/Yggdrasil/pull/10709

Would be good to release this such that this package can be used together with the latest GDAL build with HDF4 support as well.


Alexander-Barth:
Thanks a lot Martijn ! I will make a new release.

PR #288, opened 2025-09-17
move badges

alex-s-gardner:
move doc badges above others for improved discoverability


Alexander-Barth:
Good idea @alex-s-gardner ! Thanks!

PR #300, opened 2026-01-29
Ref -> Array{Bool,0} in NCDataset to make types concrete

maximilian-gelbrecht:
In the definition of `NCDataset` `Ref{Bool}` isn't a concrete type, but `Base.RefValue{Bool}` is. 

The definition with `Ref{Bool}` is causing problems for me when using NCDatasets as part of our models with Reactant.jl, and changing it to `Base.RefValue{Bool}` fixes that.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -31,7 +31,7 @@ mutable struct NCDataset{TDS,Tmaskingvalue} <: AbstractNCDataset where TDS <: Un
     iswritable::Bool
     # true of the NetCDF is in define mode (i.e. metadata can be added, but not data)
     # need to be a reference, so that remains syncronised when copied
-    # use 0-d array to as a slightly hacky solution to have a concrete type instead of Ref{Bool}
+    # use 0-d array to have a concrete type instead of a Ref{Bool} or type parameter
     isdefmode::Array{Bool, 0}
     # mapping between variables related via the bounds attribute
     # It is only used for read-only datasets to improve performance


maximilian-gelbrecht:
Thanks for the quick response!

> `foo` is not a concrete type, but a `Base.RefValue` is apparently not concrete either:

It is, you just made a small mistake there: `isconcretetype` takes in types, not instances: 

```Julia 
julia> isconcretetype(typeof(Base.RefValue(false)))
true

julia> isconcretetype(Base.RefValue{Bool})
true

julia> isconcretetype(Ref{Bool})
false
```

> Being compatible with Reactant.jl is indeed something I would like to see. :-)
I am wondering whether Base.RefValue is part of the public API ?
https://github.com/JuliaLang/julia/pull/58101

Yeah, that's a bit odd in my opinion, because why is the abstract type public API, but the concrete type isn't? 
Personally, I wouldn't mind having a `Base.RefValue` here, but it's your call! 

Btw as you seem interested about Reactant compatibility in general: so far I am just starting with this for our model (SpeedyWeather.jl). Without this fix you can't have a `NCDataset` in a struct that you use in a call to Reactant's compiler, even if you don't end up using it at all. With the fix, I am able to do that. I am not yet at actual output. But Oceananigans uses `NCDatasets` as well and is Reactant compatible, so they probably already have that figured out. The difference there is just that Oceananigans doesn't include the outputter in their model types like we do in SpeedyWeather.

Alexander-Barth:
Ah, that makes sense:

```
julia> isconcretetype(Base.RefValue{Bool})
true

julia> isconcretetype(Array{Bool,0})
true
```

While `Array{Bool,0}` seems hacky, it has minimal impact of the API and maintainability. In the future, if we are dropping or reordering type parameters, we break the API of NCDatasets, so I would like to add only type parameter if it is really necessary...
But can you check if that would be fine for Reactant.jl ?

PR #317, opened 2026-07-24
add support for zstd

bjarthur:
this depends on a new libnetcdf being compiled by binarybuild that permits plugins.  see https://github.com/JuliaPackaging/Yggdrasil/pull/13316

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -259,7 +259,7 @@ end
 
 #### Querying and Modifying Compression Settings
 
-You can query the compression settings of an existing variable, or change them on an open writeable variable, using the high-level `deflate` and `zstd` functions:
+You can query the compression settings of an existing variable, or change them on an open writeable variable, using the high-level `deflate` and `NCDatasets.zstandard` functions:
 
 ```julia
 # Open the file in append/write mode
@@ -267,11 +267,11 @@ ds = NCDataset("/tmp/zstd_example.nc", "a")
 v = ds["humidity"]
 
 # Query Zstd settings
-has_zstd, level = zstd(v)
+has_zstd, level = NCDatasets.zstandard(v)
 println("Zstd enabled: $has_zstd, level: $level")
 
 # Change Zstd compression level on the fly
-zstd(v, 10)
+NCDatasets.zstandard(v, 10)
 
 # Check Deflate settings
 isshuffled, isdeflated, deflate_level = deflate(v)
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -48,6 +48,7 @@ For more details please see the individual pages of the documentation.
 * [Load a netCDF file](@ref)
 * [Create a netCDF file](@ref)
 * [Edit an existing netCDF file](@ref)
+* [Compression (Deflate and Zstandard)](@ref)
 * [Create a netCDF file using the metadata of an existing netCDF file as template](@ref)
 * [Get one or several variables by specifying the value of an attribute](@ref)
 * [Load a file with unknown structure](@ref)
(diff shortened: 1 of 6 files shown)


bjarthur:
@mkitti

Alexander-Barth:
Thank you for nice PR!
I would propose to call these functions 
zstd(v::Variable, level::Integer)
zstd(v::Variable)

zstandard  (similar to the C api function) instead of zstd and not exporting them (but still be marking them as public) to reduce the risk of conflicts.
Do you agree?

bjarthur:
that's fine by me and i've pushed a commit with those changes.

one thing to note is that there is now an asymmetry in the API:  `deflate` is exported but `zstandard` is not.

bjarthur:
and, there are only a handful of names declared as `public`, to not include `deflate`:  https://github.com/JuliaGeo/NCDatasets.jl/blob/main/src/NCDatasets.jl#L91.  (is `eval` really needed here?)

Alexander-Barth:
> deflate is exported but zstandard is not.

yes, I am more and more coming to the conclusion that exporting names is bad for the ecosystem unless it a name that is really exclusive to a package. Unfortunately we cannot un-export `deflate` without breaking code (adding an export later is always possible).

I still want to support julia lts (1.10). Once the lts version has `public` statement, this `eval` becomes unnecessary.
My understanding is that all names that are exported are automatically public.

bjarthur:
i found a couple of remaining references to `zstd`, changed them to `zstandard`, and added a test for that section of the code, which copies a variable.  but the tests fail.  can you please take a look?  i can't figure it out.  see the latest commit.

Alexander-Barth:
The codecov tests fail because some small reduction in code coverage. But all other tests are ok.
Let me know if it can be merged for you (the github ui shows that the PR is in draft stage)

Thanks a lot!

bjarthur:
the zstd tests only pass because they are skipped if libnetcdf does not support zstd, which it currently does not until my comments [here](https://github.com/JuliaPackaging/Yggdrasil/pull/13316) are addressed.  i'd suggest waiting to merge until a new libnetcdf is available on yggdrasil

mkitti:
I just want to note that for HDF5.jl and HDF5_jll.jl we do not compile compression plugins. We implement them in Julia:
https://github.com/JuliaIO/HDF5.jl/tree/master/filters%2FH5Zzstd
https://github.com/JuliaIO/HDF5.jl/blob/master/ext%2FCodecZstdExt%2FCodecZstdExt.jl

bjarthur:
oof, ok, will look into CodeZStd.jl.  what are the pros and cons?

mkitti:
The C plugin is just 123 lines of code:
https://github.com/HDFGroup/hdf5_plugins/blob/master/ZSTD/src/H5Zzstd.c

It just glues the HDF5 API to the Zstandard API. This is Julia, so we can just translate that to Julia instead of having to compile to small shim for every platform.

Also note @nhz2 has been working on https://github.com/JuliaIO/ChunkCodecs.jl/tree/main/LibZstd

The basic question is if the underlying HDF5 APIs are accessible or if I need to use NetCDF APIs to register a plugin.

mkitti:
Just load H5Zzstd.jl:

```julia
julia> using NCDatasets

julia> mktemp() do filename, io
           mode = NCDatasets.NC_CLOBBER | NCDatasets.NC_NETCDF4
           ncid = NCDatasets.nc_create(filename, mode)
           result = NCDatasets.nc_inq_filter_avail(ncid, 32015)
           NCDatasets.nc_close(ncid)
           return result
       end
false

julia> using H5Zzstd

julia> mktemp() do filename, io
           mode = NCDatasets.NC_CLOBBER | NCDatasets.NC_NETCDF4
           ncid = NCDatasets.nc_create(filename, mode)
           result = NCDatasets.nc_inq_filter_avail(ncid, 32015)
           NCDatasets.nc_close(ncid)
           return result
       end
true
```

PR #318, opened 2026-09-17
Zstandard compression via the HDF5 filter (zstdlevel, zstandard)

bjarthur:
Adds Zstandard (zstd) compression, as an alternative to #317 with the approach @mkitti suggested there: no recompiled NetCDF_jll is needed. The zstd HDF5 filter (id 32015) is registered in-process by loading `H5Zzstd` (HDF5.jl), and since NetCDF_jll links the same `HDF5_jll` libhdf5, libnetcdf sees it.

```julia
using NCDatasets, H5Zzstd
NCDataset("f.nc", "c") do ds
    defDim(ds, "lon", 100); defDim(ds, "lat", 100)
    v = defVar(ds, "T", Float32, ("lon", "lat"); chunksizes = (50, 50), zstdlevel = 5)
    v[:, :] = rand(Float32, 100, 100)
end
```

### API
- `defVar(...; zstdlevel = n)` keyword, parallel to `deflatelevel`
- `NCDatasets.zstandard(v, level)` / `iszstd, level = NCDatasets.zstandard(v)` — public but not exported, as proposed in #317
- settings are preserved by `defVar(dest, srcvar)`

### Implementation notes
- Built on the generic `nc_def_var_filter` / `nc_inq_var_filter_info` (new wrappers) rather than `nc_def_var_zstandard`: on a libnetcdf built without zstd the latter is a silent no-op returning `NC_NOERR` (netcdf-c `libdispatch/dfilter.c`), which would write uncompressed data without warning.
- If the filter is not available, `defVar`/`zstandard` throw an error explaining to load `H5Zzstd` (or use `HDF5_PLUGIN_PATH`), instead of the raw `NetCDF: Filter error: undefined filter encountered`.
- Copying a `CFVariable` dispatches to CommonDataModel's generic `defVar`, which knows nothing about zstd; NCDatasets' copier now also accepts `CFVariable{<:Any,<:Any,<:AbstractNCVariable}` so the zstd settings survive copies. (This also affected #317.)
- Tests load `H5Zzstd` so the zstd path really runs on CI (in #317 it was skipped whenever the filter was unavailable, i.e. always).

### Caveats (documented)
- Reading also needs the filter loaded; other tools need a zstd HDF5 plugin (`HDF5_PLUGIN_PATH`, Python `hdf5plugin`).
- Negative zstd levels are accepted by libnetcdf but the current Julia filter (`H5Zzstd` 0.1.x, and HDF5.jl master's `CodecZstdExt`) fails on them (`Cint(cd_values[1])` overflow); a fix to HDF5.jl is being prepared. Tests use positive levels for now.

### Open question for the maintainers
The `using H5Zzstd` requirement (and the HDF5.jl 0.17 vs 0.18 naming churn: `H5Zzstd` vs `HDF5`+`CodecZstd`) could be removed entirely: NCDatasets already has `HDF5_jll` and `Zstd_jll` in its dependency tree, so a ~70-line filter callback backed by `ChunkCodecLibZstd` (a small, stable, HDF5.jl-independent package) registered through `HDF5_jll` directly makes plain `using NCDatasets` sufficient and supports negative levels. A prototype works against NetCDF_jll 401.1000.101 / HDF5_jll 2.2.2. The cost is two small direct dependencies (`ChunkCodecCore`, `ChunkCodecLibZstd`). Happy to switch this PR to that if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bjarthur:
@mkitti

